PR #1, opened 2022-09-06
[Feature] R client

WardBrian:
Copying over from https://gitlab.com/roualdes/bridgestan/-/merge_requests/5

This adds an R client which uses an R6 class and the .C foreign function interface.
This required shim functions to be added to bridgestan.cpp, since the .C interface requires all arguments be passed as pointers and all return values be handled by out parameters.
Closes [#8](https://github.com/roualdes/bridgestan/-/issues/8)


bob-carpenter (approved):
Thanks.  Looks good.

PR #9, opened 2022-09-08
Test threading in Julia CI

WardBrian:


src/bridgestan.cpp
```diff
@@ -327,9 +327,6 @@ stan::model::model_base& new_model(stan::io::var_context &data_context,
                                    unsigned int seed, std::ostream *msg_stream);
 
 model_rng* construct_impl(char* data_file, unsigned int seed, unsigned int chain_id) {
-  // enforce math lib thread locality for multi-threading
-  static thread_local stan::math::ChainableStack dummy;
```
roualdes, line 331: What's the difference between putting this line here and putting it in log_gradient_density_impl?

If it's better to go closer to where we call stan::math::gradient, then it probably should be added to log_density_impl, too, ya?

julia/src/Bridgestan.jl
```diff
@@ -36,9 +36,10 @@ mutable struct StanModel
 
         function f(sm)
             ccall(Libc.Libdl.dlsym(sm.lib, "destruct"),
-                  UInt32,
-                  (Ptr{StanModelStruct},),
-                  sm.stanmodel)
+                UInt32,
+                (Ptr{StanModelStruct},),
+                sm.stanmodel)
+            @async Libc.Libdl.dlclose(sm.lib)
```
roualdes, line 42: Nice catch.

roualdes (changes requested):
Just one main question on the ChainableStack variable. Otherwise LGTM.

src/bridgestan.cpp
WardBrian: I need to check that something I changed later wasn’t the “true” fix, but at first I was just trying to get the code closer to how it was in the original interface for this. I believe it might matter though, since I think the version in the constructor will have dropped out of scope by the time gradient gets called. 

If that theory is correct, you’re right about the log_density version, so something needs to change anyway

WardBrian: It was answer "B": it does need to be in the function with the gradient call, so I added it to log_density_impl as well.

PR #10, opened 2022-09-08
contrib guide, formatters, licenses, C++ std

bob-carpenter:
This PR has a bunch of things around doc and formatting:

* add `CONTRIBUTING.md` file
* moves `LICENSE.md` to `LICENSE-CODE` and add `LICENSE-DOC` with CC BY 4.0 (this will put both licenses on the pages)
* adds C++1y standard in the Makefile to match Stan
* adds .clang-format and .pylintrc format specifications

**Warning**:  I haven't actually tried to run with the new lib spec or checked that the licenses will render correctly.  I did check links.


Makefile
```diff
@@ -22,6 +22,13 @@ CXX_TYPE ?= clang
 -include $(MATH)make/dependencies
 -include $(MATH)make/libraries
 
+## set compiler standard to C++1y to match Stan
```
WardBrian, line 25: Why are these being added to the STANCFLAGS? Shouldn't it be in CXXFLAGS?

When I run `make` on a model, it already has `-std=c++1y`, I believe being pulled in from the CmdStan makefiles

bob-carpenter: Thanks.  You're right that it should've been `CXXFLAGS`.  But if it's being picked up from CmdStan, then that's even better as it'll keep up when they bump.  So I'll just remove this change.

CONTRIBUTING.md
```diff
@@ -0,0 +1,96 @@
...
+* We try to follow the [Google C++ Style Guide](https://google.github.io/styleguide/cppguide.html), but (a) we allow C++ exceptions, and (b) we allow reference arguments.
+
+* We will use [GoogleTest](https://google.github.io/googletest/) for C++ unit testing.
+
+* We recommend using [Clang format](https://clang.llvm.org/docs/ClangFormat.html) with our config file [`.clang-format`](https://github.com/roualdes/bridgestan/blob/main/.clang-format).
+
+
+### Python development
+
+* Python development relies on the external dependency:
```
WardBrian, line 67:
```suggestion
* Python development relies on the external dependencies:
```

CONTRIBUTING.md
```diff
@@ -0,0 +1,96 @@
...
+* We autoformat code with [black](https://black.readthedocs.io/en/stable/).
+
+* We recommend but do not require checking the code style and semantics with [PyLint](https://www.pylint.org) using [BridgeStan .pylintrc](https://github.com/roualdes/bridgestan/blob/main/.pylintrc).
+
+### R development
+
+* R dependencies beyond those included in this repo:
+    * [R6 package](https://cran.r-project.org/web/packages/R6/index.html) for reference classes
+    * [testthat](https://testthat.r-lib.org) for unit testing
+
```
WardBrian, line 84: I would add a note here something like

> We use the most basic [`.C` interface](https://cran.r-project.org/doc/manuals/r-release/R-exts.html#Interface-functions-_002eC-and-_002eFortran) for calling C from R.

CONTRIBUTING.md
```diff
@@ -0,0 +1,96 @@
...
+## Language specific practices
+
+### C and C++ development
+
+* We require the following dependency for C++ development:
+
+* We use the C++1y standard for compilation (`-std=c++1y` in both clang and gcc).  This is partway between C++11 and C++14, and is what Stan requires.
+
+* We try to write standards-compliant code that does not depend on features of specific platforms (except where needed for compatibility).
```
WardBrian, line 56: I think expanding this with an example is good. 

> For example, we avoid having our C++ code depend on the `R.h` or `Python.h` headers, but adding new signatures to work with a specific language's style of foreign function interface is permitted.

CONTRIBUTING.md
```diff
@@ -0,0 +1,96 @@
...
+* R dependencies beyond those included in this repo:
+    * [R6 package](https://cran.r-project.org/web/packages/R6/index.html) for reference classes
+    * [testthat](https://testthat.r-lib.org) for unit testing
+
+### Julia development
+
+* Julia dependencies beyond those included in this repo:
+    * [Test](https://docs.julialang.org/en/v1/stdlib/Test/)
+    * [LinearAlgebra](https://docs.julialang.org/en/v1/stdlib/LinearAlgebra/)
+
```
WardBrian, line 90: If you want to link to the relevant Julia docs on the C FFI, they are https://docs.julialang.org/en/v1/base/c/ and https://docs.julialang.org/en/v1/stdlib/Libdl/

WardBrian (approved):
A few small comments but otherwise LGTM

PR #12, opened 2022-09-09
[Julia] Interface Improvements

### LINKED ISSUE #11: Julia interface improvements
Here's a list of Julia interface items I'd like to tackle one day.

* adopt more idiomatic package name spelling, BridgeStan instead of Bridgestan
* export K means nothing since K is never defined
* its common Julia practice to use f! when there's an out argument; add log_density, log_density_gradient (no exclamation points) which assign to new memory (instead of using an out argument)
* can drop the ::Vector{Float64} in most use cases of zeros(D); Float64 by default
* not all defined functions are exported
* double check tests for completeness

WardBrian:
Closes #11

### DIFF
--- a/julia/src/BridgeStan.jl
+++ b/julia/src/BridgeStan.jl
@@ -29,16 +29,27 @@ mutable struct StanModel
     const seed::UInt32
     const chain_id::UInt32
 
-    function StanModel(stanlib_::String, datafile_::String, seed_ = 204, chain_id_ = 0)
+    function StanModel(stanlib_::String, datafile_::String = "", seed_ = 204, chain_id_ = 0)
         seed = convert(UInt32, seed_)
         chain_id = convert(UInt32, chain_id_)
+
+        if !isfile(stanlib_)
+            throw(SystemError("Dynamic library file not found"))
+        end
+
+        if datafile_ != ""  && !isfile(datafile_)
+            throw(SystemError("Data file not found"))
+        end
+
         lib = Libc.Libdl.dlopen(stanlib_)
 
         stanmodel = ccall(Libc.Libdl.dlsym(lib, "construct"),
                           Ptr{StanModelStruct},
                           (Cstring, UInt32, UInt32),
                           datafile_, seed, chain_id)
-
+        if stanmodel == C_NULL
+            error("could not construct model RNG")
+        end
 
         sm = new(lib, stanmodel, datafile_, seed, chain_id)
 
@@ -47,7 +58,6 @@ mutable struct StanModel
                 UInt32,
                 (Ptr{StanModelStruct},),
                 sm.stanmodel)
-            @async Libc.Libdl.dlclose(sm.lib)
         end
 
         finalizer(f, sm)
@@ -95,7 +105,7 @@ end
 function param_constrain!(sm::StanModel, theta_unc, out::Vector{Float64}; include_tp=false, include_gq=false)
     dims = param_num(sm; include_tp=include_tp, include_gq=include_gq)
     if length(out) != dims
-        error("out must be same size as number of constrained parameters")
+        throw(DimensionMismatch("out must be same size as number of constrained parameters"))
     end
     rc = ccall(Libc.Libdl.dlsym(sm.lib, "param_constrain"),
                Cint,
@@ -117,7 +127,7 @@ end
 function param_unconstrain!(sm::StanModel, theta, out::Vector{Float64})
     dims = param_unc_num(sm)
     if length(out) != dims
-        error("out must be same size as number of unconstrained parameters")
+        throw(DimensionMismatch("out must be same size as number of unconstrained parameters"))
     end
 
     rc = ccall(Libc.Libdl.dlsym(sm.lib, "param_unconstrain"),
@@ -139,7 +149,7 @@ end
 function param_unconstrain_json!(sm::StanModel, theta::String, out::Vector{Float64})
     dims = param_unc_num(sm)
     if length(out) != dims
-        error("out must be same size as number of unconstrained parameters")
+        throw(DimensionMismatch("out must be same size as number of unconstrained parameters"))
     end
 
     rc = ccall(Libc.Libdl.dlsym(sm.lib, "param_unconstrain_json"),
@@ -175,7 +185,7 @@ function log_density_gradient!(sm::StanModel, q, out::Vector{Float64}; propto = 
     lp = Ref{Float64}(0.0)
     dims = param_unc_num(sm)
     if length(out) != dims
-        error("out must be same size as number of unconstrained parameters")
+        throw(DimensionMismatch("out must be same size as number of unconstrained parameters"))
     end
 
     rc = ccall(Libc.Libdl.dlsym(sm.lib, "log_density_gradient"),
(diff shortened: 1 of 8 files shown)


julia/src/BridgeStan.jl
```diff
@@ -0,0 +1,336 @@
...
+        (Ptr{StanModelStruct},),
+        sm.stanmodel,
+    )
+    [string(s) for s in split(unsafe_string(cstr), ',')]
+end
+
+function param_constrain!(
+    sm::StanModel,
+    theta_unc,
+    out::Vector{Float64};
```
roualdes, line 126: I honestly don't know what to type `theta_unc` and `out` as, or if typing them at all is best.

The issue is that both `theta_unc` and `out` can be any type `T` such that `T` is a subtype of `AbstractVector`, with elements of `Float64`, *and* the elements are contiguous in memory.  It's the contiguous in memory part that I don't know how to type.  The rest of it would be

```
function param_constrain!(sm::StanModel, theta_unc::T, out::T; ...) where {T <: AbstractVector{Float64}}
...
end
```

Thoughts?

julia/src/BridgeStan.jl
```diff
@@ -0,0 +1,336 @@
...
+        (Ptr{StanModelStruct}, Cint, Cint, Ref{Cdouble}, Ref{Cdouble}),
+        sm.stanmodel,
+        include_tp,
+        include_gq,
+        theta_unc,
+        out,
+    )
+    if rc != 0
+        error("param_constrain failed on C++ side; see stderr for messages")
+    else
```
roualdes, line 148: Change or not, as this is definitely a personal preference.

I'd prefer to end the if and return immediately after, instead of returning from within the else block.

julia/src/BridgeStan.jl
```diff
@@ -0,0 +1,336 @@
...
+    end
+end
+
+function param_unconstrain_json(sm::StanModel, theta::String)
+    out = zeros(param_unc_num(sm))
+    param_unconstrain_json!(sm, theta, out)
+end
+
+function log_density(sm::StanModel, q; propto = true, jacobian = true)
+    lp = Ref{Float64}(0.0)
```
roualdes, line 220:
```suggestion
    lp = Ref(0.0)
```
The number `0.0` makes the type `Float64` implicitly.

julia/src/BridgeStan.jl
```diff
@@ -0,0 +1,336 @@
...
+end
+
+function log_density_gradient!(
+    sm::StanModel,
+    q,
+    out::Vector{Float64};
+    propto = true,
+    jacobian = true,
+)
+    lp = Ref{Float64}(0.0)
```
roualdes, line 245:
```suggestion
    lp = Ref(0.0)
```

julia/src/BridgeStan.jl
roualdes: Whichever you pick, will you make it consistent throughout?

julia/src/BridgeStan.jl
```diff
@@ -0,0 +1,336 @@
...
+
+function log_density_hessian!(
+    sm::StanModel,
+    q,
+    out_grad::Vector{Float64},
+    out_hess::Vector{Float64};
+    propto = true,
+    jacobian = true,
+)
+    lp = Ref{Float64}(0.0)
```
roualdes, line 286:
```suggestion
    lp = Ref(0.0)
```

roualdes (commented):
Overall this looks great.  Just a few small suggestions and one main question about typing some arguments (I asked the question at one function, but the ideas will carry over to other functions).

What's going on with the edit on the `python/test/test_bridgestan.py` file?

julia/src/BridgeStan.jl
```diff
@@ -0,0 +1,336 @@
...
+
+        sm = new(lib, stanmodel, datafile_, seed, chain_id)
+
+        function f(sm)
+            ccall(
+                Libc.Libdl.dlsym(sm.lib, "destruct"),
+                UInt32,
+                (Ptr{StanModelStruct},),
+                sm.stanmodel,
+            )
```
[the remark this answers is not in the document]
WardBrian: This actually causes a segfault in the tests if added. I believe the Julia runtime sometimes closes the DSO before the async part gets run. Closer reading of the doc suggests only using @async in finalizers for debugging, so I don’t think doing something like this is fully supported.

julia/src/BridgeStan.jl
```diff
@@ -0,0 +1,336 @@
...
+end
+
+function param_unc_names(sm::StanModel)
+    cstr = ccall(
+        Libc.Libdl.dlsym(sm.lib, "param_unc_names"),
+        Cstring,
+        (Ptr{StanModelStruct},),
+        sm.stanmodel,
+    )
+    [string(s) for s in split(unsafe_string(cstr), ',')]
```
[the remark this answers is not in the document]
WardBrian: I don't think I had really grocked the `.` semantics until midway through writing the test code. Super cool!

julia/src/BridgeStan.jl
WardBrian: I'm not that familiar with Julia's more exotic types, can you give an example of something which is `<: AbstractVector{Float64}`, has contiguous memory, but is _not_ `Vector{Float64}`? I'd like to include a test with that thing if we expand this type.

This is also the kind of thing which we could change in a backwards-compatible way later on

roualdes: Yep, you're right.  Being stricter now, by forcing `Vector{Float64}`, will allow us to accept a broader set of types in the future in a backwards-compatible way, if we need/want to.  Good point.


For now we still want to explicitly avoid something like 

```
M = rand(3, 3);
@views M[1, :] # not contiguous
```
being passed in.

Will you add the type `Vector{Float64}` to the arguments `theta_unc`, `theta`, and `q`? I think all `out` arguments have it already.

Views, and less important for our use case ranges, are both `<: AbstractVector{Float64}`, have contiguous memory, and are *not* `Vector{Float64}`.  Try

```
x = rand(10);
typeof(@views x[1:2]) <: AbstractVector{Float64}
typeof(@views x[1:2])
y = 1.0:10.0;
typeof(y) <: AbstractVector{Float64}
typeof(y)
```

And for future reference, here's a [discussion of this issue](https://github.com/JuliaLang/julia/issues/10889) with a [temporary solution](https://github.com/JuliaIO/Blosc.jl/blob/25d663c607542cbebaea45542619726fec71bb5e/src/Blosc.jl#L58).

roualdes:
Any thoughts on the edit of the python/test/test_bridgestan.py file?

WardBrian:
test_bridgestan.py: While copying the tests over to Julia I noticed a copy/paste error in the test. The chunk I deleted is an exact duplicate of the chunk above it, see https://github.com/roualdes/bridgestan/blob/8e2c286ed7bf94a406ebf07af0b0cf707762754f/python/test/test_bridgestan.py#L348-L367

julia/src/BridgeStan.jl
WardBrian: As was just revealed to me by a test that broke when fixing those types, it seems that Julia also does the right thing when given a `Matrix{Float64}` to a C call. So really, down the road, we don't even need AbstractVector, but just AbstractArray and continuous?

PR #13, opened 2022-09-12
Standardize class names in interfaces

WardBrian:
While working on #12 I noticed that Julia named the type exposed by the module `StanModel`, where as the Python and R interfaces called it `Bridge`. 

I personally think `StanModel` is a better name for what the conceptual object being created is, so this PR renames it in the Python and R interfaces.

### DIFF
--- a/R/bridgestan.R
+++ b/R/bridgestan.R
@@ -1,5 +1,5 @@
 
-Bridge <- R6::R6Class("Bridge",
+StanModel <- R6::R6Class("StanModel",
   public = list(
     initialize = function(lib, data, rng_seed, chain_id) {
       if (.Platform$OS.type == "windows"){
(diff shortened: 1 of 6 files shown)


bob-carpenter:
What you're calling `StanModel` includes the model and some precompiled metadata, which are immutable, along with a mutable RNG.  So I was reluctant to call it `StanModel`.

I've been trying to work through some Python threading material, and this is the kind of thing where multiple threads would use different RNG states.

bob-carpenter (approved):
looks good assuming we want to go with the `StanModel` name.  I'm OK with that, but it misses the RNG component, which is mutable, unlike the model and its metadata.

roualdes:
I both agree that StanModel isn't perfect and that StanModel will suit users better than Bridge.

Also happy to consider other names.

bob-carpenter:
I believe in merging things that improve the code base, so I merged this.  If we come up with a better name before release, we can change again.

Issue #7, opened 2022-09-08
Interfaces doc: sphinx hybrid or other solution?


roualdes:
I naively expected Sphinx to provide well means to doc other languages than Python.  With a bit more research, I'm losing my optimism.  

Brute forcing Julia doc into Sphinx looked atrocious. 

There's a [Julia domain](https://bastikr.github.io/sphinx-julia/juliadomain.html), which looks like it could be tricky to get right in a GitHub Action.  I'll try it out locally tomorrow and report back.

I have not found a solution for R.

Here's a hybrid idea.  We doc each interface using the common, or at least a preferred, tool within that particular language to produce an HTML file.  Include the HTML file in the static directory of the sphinx pages, and then provide hyperlinks to those pages.

Pros: it should work; all interfaces would have doc in a format for which its users would better recognize.

Cons: all interface docs would look different; the seamless "Next >" buttons on the bottom right of the pages would make less sense if we hyperlink out of the languages folder.

Any other ideas?

WardBrian:
The Julia domain wouldn't be that hard to get working in Github Actions... I just can't get it to work _period_. I suspect it was developed using a prior major version of Sphinx and doesn't seem like it was updated. The version on pypi doesn't run at all, and the version on github fails for me when I try to actually document anything. 

For Julia at least, I think the best thing to do might be to combine https://github.com/JuliaDocs/DocumenterMarkdown.jl and Sphinx extension [MyST](https://www.sphinx-doc.org/en/master/usage/markdown.html). Unless this is the atrocious output you described.

roualdes:
Good thought.  Didn't know about such options.

Unfortunately, the DocumenterMarkdown.jl README suggests development is less than active, and it only works with the previously released version of Documenter.jl.

roualdes:
Solved by PRs #15, #16, and #20:  Sphinx hybrid.

PR #15, opened 2022-09-14
Add Julia doc and automated Documenter -> Sphinx pipeline

WardBrian:
This was discussed in the comments of #7.

This uses `DocumenterMarkdown` to build a Markdown version of the Julia docs and embeds it in Sphinx. The output is not quite as pretty as Python's, but I think it's not too bad:

image

If you don't have Julia installed, I am assuming you're not updating the Julia docs, so it will ignore the error and proceed.

### DIFF
--- a/.github/workflows/docs.yaml
+++ b/.github/workflows/docs.yaml
@@ -32,12 +32,15 @@ jobs:
       - name: Install dependencies (python)
         run: |
           python -m pip install --upgrade pip wheel
-          python -m pip install --upgrade "sphinx>5" nbsphinx ipython ipykernel ipywidgets pydata-sphinx-theme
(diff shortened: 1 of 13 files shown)


WardBrian:
Tested that the GHA build still works on my fork. As a bonus, that means we can preview the output: https://brianward.dev/bridgestan/languages/julia.html

WardBrian:
Yeah, I'm not sure how best to communicate this. The file `julia/doc/src/julia.md` is actually the source file. The file `python/doc/languages/julia.md` gets deleted each re-build. I wanted to include a comment in the resulting file to warn people, but of course that comment must also appear in the source. 

Should I just try to reword it?

PR #17, opened 2022-09-16
Docs

rok-cesnovar:
@WardBrian I just added a few lines to Description and then just used `devtools::document()`.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -26,3 +26,4 @@ bridgestan.Rcheck/
 build/
 
 .DS_Store
+.Rproj.user
--- a/R/DESCRIPTION
+++ b/R/DESCRIPTION
@@ -7,9 +7,12 @@ License: BSD_3_clause
 Description: BridgeStan provides efficient in-memory access to the methods of a Stan model,
     including log densities, gradients, Hessians, and constraining and unconstraining
     transforms.
+Encoding: UTF-8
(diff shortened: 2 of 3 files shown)


WardBrian:
Thanks! I want to check that the extra text processing that markdown=TRUE does doesn’t break the Rd2txt conversion we use for generating a docs web page. 

It’s not a huge deal if it does necessarily, since that’s already a manual process, and going forward I think a lot of the updates will need to be copied over manually rather than automatically for this interface

rok-cesnovar:
There is also a way to automatically generate a docs website, like we have for cmdstanr for example: https://mc-stan.org/cmdstanr/reference/index.html using pkgdown. Its really easy and fast.

rok-cesnovar:
What I typically use in package development is to just run devtools::document() in GitHub Actions and push the changes to the branch. And that is all.

rok-cesnovar:
Ah yes, that makes sense! In general, the docs and package look good though. I just opened it in Rstudio, built it, and all the docs look good.

WardBrian:
Great to hear! I was doing this all without RStudio (and for a while, without using `usethis`, which seems to be the only thing tutorials recommend) so it was more manual than I would have liked

rok-cesnovar:
Yeah, without usethis you would be considered a boomeR :)

PR #20, opened 2022-09-16
Add C++ API to docs

WardBrian:
While looking up ways to hack doxygen to give me a markdown output, I discovered something way better: [breathe](https://breathe.readthedocs.io/en/stable/index.html), which is a native sphinx-doxygen interop package.

I did this up in a similar fashion as the Julia docs; when building locally, it will gracefully skip this part if you don't have the prerequisites installed. 

Preview: https://brianward.dev/bridgestan/languages/cpp-api.html


roualdes:
This is great.   I'm all for this.  

Should/can we add something to the C++ API page that explains a bit more of what role these functions serve?  On the "Next > How It Works" page we talk about the C-interface, which, I imagine, might cause confusion with the C++ API page.

WardBrian:
Yep, trying to clean up the output a bit still. Trying to grok Doxygen groups and how they render here

WardBrian:
@roualdes how's the latest changes? Preview should have re-built as well

roualdes:
Ya, this looks great.  The added structure and the extra explanation is helpful.

In the preview, the link to How it Works from within the first paragraph does not have the correct destination URL.  Is this a preview issue that won't carry over to the BridgeStan docs or a typo?

WardBrian:
That’s a real typo, my bad

WardBrian:
I’m considering hiding the *_impl functions in the doc, since they’re really not part of the public API of the server. Besides that this should be ready to go

bob-carpenter:
This looks really great.  I just defined this all as a `struct`, so everything's public.  The only thing that should be public is the constructor, destructor, and methods.  I'll open a PR to give the C++ a little privacy.

bob-carpenter:
Oops, forgot this is all in a very plain C wrapper.  Even in C++, public vs. private is just for class members.

A common convention is to use an `internal` namespace for functions you want to warn users away from.

I'm going to have to clean up the `model_rng` class by making all those member variables private and just defining the `_impl` functions as class methods.  I think that should work with C as it doesn't change the memory layout of the compiled class.

I think we might also be able to pull out a C header file (with `.h` suffix) that declares all the C functions.  Clients would then include that header and link against a compiled object file.

I'll see if I can make it work.

WardBrian:
Actually enforcing any sort of privacy seems unnecessary to be, since the _impl methods are not under the C linkage it would be very difficult for anyone to try to call them. A .h library would be very cool, but I think we’d need to make the model_rng argument a void pointer then.

It’s very easy to hide these things in this doc, we just need to remove “public-attributes” and “funcs” from the sections list and they should be off the page

WardBrian:
After #21, I've updated this to reflect the new API being defined in `bridgestan.h`. See the new output here:
https://brianward.dev/bridgestan/languages/c-api.html

I believe this is ready to merge, if either @bob-carpenter or @roualdes want to give it a once over

roualdes:
Sure, I'll give it a quick review.

python/docs/conf.py
```diff
@@ -73,21 +73,44 @@
...
+RUNNING_IN_CI = os.environ.get("CI") or os.environ.get("READTHEDOCS")
+
 try:
     print("Building Julia doc")
-    subprocess.run(
-        ["julia", '--project=.', "./make.jl"],
-        cwd=pathlib.Path(__file__).parent.parent.parent / "julia" / "docs",
-        check=True,
-    )
+    # subprocess.run(
```
roualdes, line 92: Will you remind me why this is commented out?

src/bridgestan.h
```diff
@@ -219,7 +219,7 @@ int log_density_gradient(model_rng* mr, bool propto, bool jacobian,
  * @param[in] propto `true` to discard constant terms
  * @param[in] jacobian `true` to include change-of-variables terms
  * @param[in] theta unconstrained parameters
- * @param[out] lp log density to be set
+ * @param[out] val log density to be set
```
roualdes, line 222: Good catch to align doc with the function. 

Isn't it `lp` in all other functions?  Anyway, this is something to double check later.

roualdes (approved):
Looks good to me.

Does my review prevent a second review?

python/docs/conf.py
WardBrian: My mistake when trying to iterate quicker on the docs

src/bridgestan.h
WardBrian: It's called `lp` for log_density and `val` for the gradient and Hessian functions. Agreed would be good to do a pass on later, for now it was just generating a doxygen warning

PR #21, opened 2022-09-18
factored into headers

bob-carpenter:
This is a big refactor on the C/C++ side.   Everything should work as before if you comple `bridgestan.cpp`. Going forward, I'd like to factor the builds a bit so we can build `bridgestan.cpp` without `bridgestanR.cpp`.

Aside from moving things around, the biggest change is pulling all the implementation logic into the `model_rng` class.  We should eventually convert `bridgeStanR.cpp` to calling the member functions rather than the `bridgestan.h` functions.  Then `bridgestan.cpp` and `bridgestanR.cpp` will be independent rather than having the latter depend on the former.

Most importantly, I pulled implementations into the C++ class, made all the members private, and removed the `_impl` functions.  This should make the doc a lot cleaner. I then pulled out relevant headers, including putting the R header in its own file.  Here's where things went:

#### `model_rng.h`
*Declares:*  class `model_rng`
*Depends on:* `<stan/model/model_base.hpp>`,  `<boost/random/additive_combine.hpp>`, `<string>`, and `<vector>`

#### `bridgestan.h`
*Declares:* simple external C interface
*Depends on:* `model_rng.h`

#### `bridgestanR.h`
*Declares:*:  external C interface for R
*Depends on*: `model_rng.h`

#### `model_functor.hpp`
*Declares:* `model_functor`
*Depends on:* `<Eigen/Dense>`, `<ostream>`

#### `model_rng.cpp`
*Defines:* `model_rng.h`
*Depends on:*` everything

#### `bridgestanR.cpp`
*Defines:* `bridgestanR.h`
*Depends on*: `bridgestan.cpp`

#### `bridgestan.cpp`
*Defines*: `bridgestan.h`
*Depends on:* `model_functor.hpp`, `model_rng.cpp`, `bridgestanR.cpp`


WardBrian:
Can I ask why .h instead of .hpp? I don’t believe these headers would be valid to include in a C file as is. Is that a goal of this? (I guess a better question is: what are the goals of this?)

The docs also should be a bit cleaner even before this PR, see https://brianward.dev/bridgestan/languages/cpp-api.html for the latest preview which hid all the _impl and members of the struct

bob-carpenter:
The goals are to

1.  structure the code into headers and implementation files so clients can include headers they need and we can link in object files,
2.  break up the enormous files across clean module lines to minimize dependencies, and
3. make private things private.

This is the standard way to organize C and C++ interfaces---separating headers and definitions (header-only C++ makes it possible to just include everything, but it can lead to huge compile times) so that there can be one-way includes from bigger to smaller modules.  We should be able to precompile everything here other than the model class .hpp file.

As to why `.hpp` rather than `.h`, I was thinking from the client side, where Including `.hpp` seems strange from C.  From the code side, it'd make sense to call it `.hpp`.  I'm not wedded to making it `.h`, so we can change it if you'd prefer.  So I left the thing only called from C++ as a header-only `.hpp` file.

bob-carpenter:
How do I run the Python tests?  I thought I just needed to call `pytest` but that gives me this puzzling error dump.

```
~/github/roualdes/bridgestan/python(feature/cpp-privacy)$ pytest
=================================== test session starts ====================================
platform darwin -- Python 3.9.9, pytest-6.2.5, py-1.11.0, pluggy-1.0.0
rootdir: /Users/carp/github/roualdes/bridgestan/python
collected 0 items / 1 error                                                                

========================================== ERRORS ==========================================
_________________________ ERROR collecting test/test_bridgestan.py _________________________
ImportError while importing test module '/Users/carp/github/roualdes/bridgestan/python/test/test_bridgestan.py'.
Hint: make sure your test modules/packages have valid Python names.
Traceback:
/usr/local/Cellar/python@3.9/3.9.9/Frameworks/Python.framework/Versions/3.9/lib/python3.9/importlib/__init__.py:127: in import_module
    return _bootstrap._gcd_import(name[level:], package, level)
test/test_bridgestan.py:4: in <module>
    import bridgestan as bs
E   ModuleNotFoundError: No module named 'bridgestan'
================================= short test summary info ==================================
ERROR test/test_bridgestan.py
!!!!!!!!!!!!!!!!!!!!!!!!!! Interrupted: 1 error during collection !!!!!!!!!!!!!!!!!!!!!!!!!!
===================================== 1 error in 0.15s =====================================
```

WardBrian:
For Pytest to pick up bridgestan you need to first run `pip install .` from the python folder. If you’d like, you can add `-e` to make it so the installation is “editable”, meaning you don’t need to re-install after making changes. 

I’m on board with the idea of headers generally. I think if we want to call them .h we should make it so that the C parser could handle them, which I don’t believe it currently could. I think we need to either move the `extern “C”` part into the implementation files or surround it with `#ifdef __cplusplus`

WardBrian:
Besides guarding the `extern` portions, I think to actually have this compile in a pure-C usage we also can’t have the model_rng type exposed but instead need to use void pointers everywhere. It won’t be possible to include these in C files if it’s necessary to include the class definition as well. 

Using void pointers instead is effectively what is happening in the interfaces right now anyway, but we lose some convenience and type safety on the C++ side doing it like that.

bob-carpenter:
I didn't change the interface.

The current version uses `model_rng*`,

```cpp
model_rng* construct(char* data_file, unsigned int seed,
		       unsigned int chain_id);
```

See line 86 of https://github.com/roualdes/bridgestan/blob/main/src/bridgestan.cpp.  The C interface doesn't do anything with it, though, other than pass it around.

WardBrian:
Yes, that’s fine for current interfaces. I mean that if someone writes a pure C file and does `#include bridgestan.h` they will get parser errors. We should either fix that or not call them .h, in my opinion

bob-carpenter:
Aren't the Python and R interfaces pure C?  I guess I'm not sure what constitutes "pure C" or what the `extern "C"` block does to launder C++ naming to C.  Even if we declare a `void*` and cast internally, we're still pointing to a C++ object in some sense either way. 

We can change this back to `void*` with casting if we need to.

Any hints on how to run the Python tests?  I can't get them to run as written.

WardBrian:
I'm discussing something a bit different than you are I believe. Yes, `extern "C"` means that we expose a C-compatible interface at _runtime_. That is true whether we do it as we currently do with `model_rng*`, or with `void *`.

I'm discussing specifically the naming convention of using `.h` for C headers. This is not a runtime concern but a compile-time issue. I believe that a file which ends in `.h` should be processable by a C compiler, which these are not. 

You can try it: `file test.c`
```C
#include "bridgestan.h"


int main(){
    (void*) model = construct("foo.json", 123, 0);
    return 0;
}
```

This will fail to compile because model_rng.h is including the Stan headers which use C++ specific headers. Even if that wasn't the case, it would fail because C has no idea how to parse `extern "C"` or `class`. I believe these problems are fixable, if we want them to be, as I described above. We should either make it so a program like the above can compile, or rename them to `.hpp` to indicate that they are **not** C headers.

bob-carpenter:
I realize I have no idea how this whole setup is working.  But we can talk more in person.

Right now, I would dearly love to be able to run the Python tests, but can't figure out how to make that happen.

roualdes:
@bob-carpenter to run our Python tests, try 

```
cd <bridgestan>/python
pip install -e .
pytest test/test_bridgestan.py
```
from within a Python3 environment and with pytest installed.

stan/bernoulli/bernoulli.stan
```diff
@@ -9,7 +9,7 @@ transformed parameters {
   real logit_theta = logit(theta);
 }
 model {
-  theta ~ beta(1,1);
+  theta ~ beta(2, 2);
```
WardBrian, line 12: I assume this is why the tests are failing? We test against the returned value of log_prob for this model

bob-carpenter: That makes sense.  I can change that back.  I was trying to do minimal changes so that the flags all did something, but wound up just adding a regression example.

bob-carpenter: Right now, I'm hung up on getting the paths right for the build-all.  I'm terrible at all this dev ops stuff.

WardBrian: I think I fixed the python test paths in #22 if you merge in `main`

bob-carpenter:
@WardBrian:  After we add the ifdefs for C/C++, how do we name the file?  Or do we have two?

WardBrian:
After the ifdefs are added, to `bridgestan` and `bridgestan_R` we should call them `.h`, since they are C-compatible headers. This type of thing is pretty standard I believe, if you look at for example [GNU's source for stdio.h](https://www.gnu.org/software/m68hc11/examples/stdio_8h-source.html) you can see the same 

```c
 #ifdef __cplusplus
 extern "C" {
 #endif
```

bob-carpenter:
To finish this, I'd like to:

* add the ifdefs for C vs. C++ and change the file names (just bridgestan.h and bridgestanR.h, not model_rng.hpp as the latter is just for the C++ back end)
    *  but I don't know enough about building from C to add a test that it works from C

* move the C++ source from directory src to directory cpp so that our naming is consistent across languages

* somehow get the build-all.sh script working again with path issues---it's the last piece of automating testing I haven't been able to get working locally

* improve the doc to the point where someone like me can use the package (much more explicit instructions on paths and how to get things working---I can never figure that stuff out on my own and the combinatorics involved in the guesswork is prohibitive and frustrating, as evidenced by my stream of messages and the fact that I still can't get testing working locally)

bob-carpenter:
@WardBrian:  Can you finish off the code and builds for this?  

Then I will go through and test our doc and update it so that someone with my skill level in dev ops can use it going forward.

WardBrian:
What isn't working for you with build_all.sh? If you have `$BRIDGESTAN` defined in your shell to the top-level directory of the repo the script should base all paths off that

I also am against renaming from `src` to `cpp`, I think it's useful to think of BridgeStan itself being the C++ and the other languages all being interfaces to it

bob-carpenter:
Here's what's blocking me with the build script:

```
~/github/roualdes/bridgestan(feature/cpp-privacy)$ stan/build_all.sh 
make: *** No rule to make target `stan/throw_tp/throw_tp_model.so'.  Stop.
make: *** No rule to make target `stan/throw_gq/throw_gq_model.so'.  Stop.
make: *** No rule to make target `stan/throw_lp/throw_lp_model.so'.  Stop.
make: *** No rule to make target `stan/throw_data/throw_data_model.so'.  Stop.
make: *** No rule to make target `stan/jacobian/jacobian_model.so'.  Stop.
make: *** No rule to make target `stan/matrix/matrix_model.so'.  Stop.
make: *** No rule to make target `stan/simplex/simplex_model.so'.  Stop.
make: *** No rule to make target `stan/full/full_model.so'.  Stop.
make: *** No rule to make target `stan/stdnormal/stdnormal_model.so'.  Stop.

--- Compiling Stan bridge C++ code ---
c++    -c -fPIC  -I src/ -o src/bridgestan_threads.so  src/bridgestan.cpp
In file included from src/bridgestan.cpp:1:
In file included from src/bridgestan.hpp:4:
src/model_rng.hpp:4:10: fatal error: 'stan/model/model_base.hpp' file not found
#include <stan/model/model_base.hpp>
         ^~~~~~~~~~~~~~~~~~~~~~~~~~~
...
1 error generated.
make: *** [src/bridgestan_threads.so] Error 1
```

No luck with trying it from one level down, either.

```
~/github/roualdes/bridgestan(feature/cpp-privacy)$ cd stan
~/github/roualdes/bridgestan/stan(feature/cpp-privacy)$ ./build_all.sh 
make: *** No rule to make target `stan/throw_tp/throw_tp_model.so'.  Stop.
make: *** No rule to make target `stan/throw_gq/throw_gq_model.so'.  Stop.
make: *** No rule to make target `stan/throw_lp/throw_lp_model.so'.  Stop.
make: *** No rule to make target `stan/throw_data/throw_data_model.so'.  Stop.
make: *** No rule to make target `stan/jacobian/jacobian_model.so'.  Stop.
make: *** No rule to make target `stan/matrix/matrix_model.so'.  Stop.
make: *** No rule to make target `stan/simplex/simplex_model.so'.  Stop.
make: *** No rule to make target `stan/full/full_model.so'.  Stop.
make: *** No rule to make target `stan/stdnormal/stdnormal_model.so'.  Stop.
make: *** No rule to make target `stan/bernoulli/bernoulli_model.so'.  Stop.
make: *** No rule to make target `stan/gaussian/gaussian_model.so'.  Stop.
make: *** No rule to make target `stan/fr_gaussian/fr_gaussian_model.so'.  Stop.
make: *** No rule to make target `stan/simple/simple_model.so'.  Stop.
make: *** No rule to make target `stan/multi/multi_model.so'.  Stop.
```

bob-carpenter:
> I think it's useful to think of BridgeStan itself being the C++ and the other languages all being interfaces to it

I think of this in exactly the opposite way.  BridgeStan is the interface in Python or Julia and the C++ is just the scaffolding to make it work.  But I'm OK keeping this as source---it's just people like me who'll be puzzled as to where the C++ code is and why there is both a `src` and a `python` directory.

WardBrian:
I guess to me, if someone in Julia says "I'm using BridgeStan" and someone in Python says "I'm using BridgeStan", the only thing they really have in common is the C++ code, hence that is the "source". I'm not extremely wedded to this, but the fact that all the other folders strongly depend on the C++ folder makes it special to me

WardBrian:
> Here's what's blocking me with the build script:

if you `echo $BRIDGESTAN` does it print out anything? You might have to run `export BRIDGESTAN=$(pwd)` from the top level folder (or whatever your favorite way of setting an environment variable is)

bob-carpenter:
Still no luck.  With our without final slash (this is the kind of combinatorics that just kills me).

```
~/github/roualdes/bridgestan(feature/cpp-privacy)$ echo $BRIDGESTAN
/Users/carp/github/roualdes/bridgestan
~/github/roualdes/bridgestan(feature/cpp-privacy)$ stan/build_all.sh 
make: *** No rule to make target `stan/throw_tp/throw_tp_model.so'.  Stop.
make: *** No rule to make target `stan/throw_gq/throw_gq_model.so'.  Stop.
make: *** No rule to make target `stan/throw_lp/throw_lp_model.so'.  Stop.
make: *** No rule to make target `stan/throw_data/throw_data_model.so'.  Stop.
make: *** No rule to make target `stan/jacobian/jacobian_model.so'.  Stop.
make: *** No rule to make target `stan/matrix/matrix_model.so'.  Stop.
make: *** No rule to make target `stan/simplex/simplex_model.so'.  Stop.
make: *** No rule to make target `stan/full/full_model.so'.  Stop.
make: *** No rule to make target `stan/stdnormal/stdnormal_model.so'.  Stop.

--- Compiling Stan bridge C++ code ---
c++    -c -fPIC  -I src/ -o src/bridgestan_threads.so  src/bridgestan.cpp
In file included from src/bridgestan.cpp:1:
In file included from src/bridgestan.h:6:
src/model_rng.hpp:4:10: fatal error: 'stan/model/model_base.hpp' file not found
#include <stan/model/model_base.hpp>
         ^~~~~~~~~~~~~~~~~~~~~~~~~~~
1 error generated.
make: *** [src/bridgestan_threads.so] Error 1
make: *** No rule to make target `stan/gaussian/gaussian_model.so'.  Stop.
make: *** No rule to make target `stan/fr_gaussian/fr_gaussian_model.so'.  Stop.
make: *** No rule to make target `stan/simple/simple_model.so'.  Stop.

--- Compiling Stan bridge C++ code ---
c++    -c -fPIC  -I src/ -o src/bridgestan_threads.so  src/bridgestan.cpp
In file included from src/bridgestan.cpp:1:
In file included from src/bridgestan.h:6:
src/model_rng.hpp:4:10: fatal error: 'stan/model/model_base.hpp' file not found
#include <stan/model/model_base.hpp>
         ^~~~~~~~~~~~~~~~~~~~~~~~~~~
1 error generated.
make: *** [src/bridgestan_threads.so] Error 1
~/github/roualdes/bridgestan(feature/cpp-privacy)$ cd stan/
~/github/roualdes/bridgestan/stan(feature/cpp-privacy)$ ./build_all.sh 
make: *** No rule to make target `stan/throw_tp/throw_tp_model.so'.  Stop.
make: *** No rule to make target `stan/throw_gq/throw_gq_model.so'.  Stop.
make: *** No rule to make target `stan/throw_lp/throw_lp_model.so'.  Stop.
make: *** No rule to make target `stan/throw_data/throw_data_model.so'.  Stop.
make: *** No rule to make target `stan/jacobian/jacobian_model.so'.  Stop.
make: *** No rule to make target `stan/matrix/matrix_model.so'.  Stop.
make: *** No rule to make target `stan/simplex/simplex_model.so'.  Stop.
make: *** No rule to make target `stan/full/full_model.so'.  Stop.
make: *** No rule to make target `stan/stdnormal/stdnormal_model.so'.  Stop.

--- Compiling Stan bridge C++ code ---
c++    -c -fPIC  -I src/ -o src/bridgestan_threads.so  src/bridgestan.cpp
In file included from src/bridgestan.cpp:1:
In file included from src/bridgestan.h:6:
src/model_rng.hpp:4:10: fatal error: 'stan/model/model_base.hpp' file not found
#include <stan/model/model_base.hpp>
         ^~~~~~~~~~~~~~~~~~~~~~~~~~~
1 error generated.
make: *** [src/bridgestan_threads.so] Error 1
make: *** No rule to make target `stan/gaussian/gaussian_model.so'.  Stop.
make: *** No rule to make target `stan/fr_gaussian/fr_gaussian_model.so'.  Stop.
make: *** No rule to make target `stan/simple/simple_model.so'.  Stop.

--- Compiling Stan bridge C++ code ---
c++    -c -fPIC  -I src/ -o src/bridgestan_threads.so  src/bridgestan.cpp
In file included from src/bridgestan.cpp:1:
In file included from src/bridgestan.h:6:
src/model_rng.hpp:4:10: fatal error: 'stan/model/model_base.hpp' file not found
#include <stan/model/model_base.hpp>
         ^~~~~~~~~~~~~~~~~~~~~~~~~~~
1 error generated.
make: *** [src/bridgestan_threads.so] Error 1
```

bob-carpenter:
Anyway, getting me working on this isn't critical at the moment.  If you're willing to update everything, just get it working on your end, then when I'm back at work tomorrow we can sit down and I'll take extensive notes on what exactly is required to get this working and put it into the doc.

WardBrian:
Huh. The only other thing I can think of is in the past something like that error has shown up when cmdstan/stanc couldn't be found

At this point, how would you feel about merging this and then doing follow ons for:

1. Developer doc with clearer steps
2. Update #20 to doc the `bridgestan.h` API
3. Add a C example to the repo (probably just the same one I sent you)
4. Discuss moving `src` in an issue

bob-carpenter:
I"m OK with doing this in whatever stages you think make most sense.  I'm not worried about users until we get to a 1.0 release.  But then I'm the sort of person who will only use a 0.1 release if I'm desperate or the project uses Xeno naming and never plans to get to a 1.0.

WardBrian:
I'm also not too worried about users at this stage, more just about keeping PRs small/atomic when we can

bob-carpenter:
This is listed as my PR, so I can't review it.  But it doesn't look like we have a require-review hook in place so feel free to merge.

PR #25, opened 2022-09-19
Add example of calling from C

WardBrian:
Discussed in #21, the refactor into headers lets you call bridged models from pure C. This creates an `c-example/` folder with a minimal example of how to do this and how to compile it.

### DIFF
--- a/c-example/README.md
+++ b/c-example/README.md
@@ -5,7 +5,7 @@ This shows how one could write a C program which calls BridgeStan.
 Any compiled language with a C foreign function interface and
 the ability to link against C libraries should be able to work similarly.
 
-## Usage - dynamic linking
+## Usage with dynamic linking
 
 It is possible to link against the same `name_model.so` object used by the other
 BridgeStan interfaces. This creates a dynamic link.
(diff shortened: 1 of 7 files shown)


WardBrian:
I spent some time trying to get Windows CI for this as well. I am 99% sure it will work in real use cases. It compiles fine in CI, but running it fails as I am once again fighting against how polluted the dynamic environment for Github Actions is.

python/docs/languages/c-api.rst
```diff
@@ -7,7 +7,9 @@ Core API
 --------
 
 The following are the C functions exposed by the BridgeStan library in ``bridgestan.h``.
-These are wrapped in the various high-level interfaces.
+These are wrapped in the various high-level interfaces. An example calling
+these functions from pure-C can be found in the ``c-example/`` subdirectory
```
bob-carpenter, line 11: I don't know the .rst format.  I did try to track down how-it-works, but I don' tknow how to verify this link works, assuming the single back tick is a link.

c-example/README.md
```diff
@@ -0,0 +1,55 @@
+# BridgeStan from C
+
+This shows how one could write a C program which calls BridgeStan.
+
+Any compiled language with a C foreign function interface and
+the ability to link against C libraries should be able to work similarly.
+
+## Usage - dynamic linking
```
bob-carpenter, line 8: Usage - dynamic -> Usage with dynamic

and same for static case below.  A colon could also work.

bob-carpenter (approved):
Looks good.  I don't read Makefile natively, but I checked out your branch and followed the instructions and it worked.  Thanks for writing up an example with clear enough doc for me!

WardBrian:
Glad it worked! The fact that this is possible continues to wow me, even if it can be a bit fiddly at times

PR #27, opened 2022-09-21
Add compilation utilities to Python interface, clean up repository

### LINKED ISSUE #26: Repository naming conventions
Originally discussed in #21:
We currently have a bunch of folders named after languages (`R`, `julia`, `python`, `c-example`, `stan`) and one named `src`.

@bob-carpenter proposed renaming `src` to `cpp`. 
I'm of the opinion that `src` _is_ a special case, since all the other languages need the code contained there to function.

I do on the other hand think that `stan` sticks out, and we might want to rename it to something like `test_models` (since that's what they are). If we do that. then we have the convention that `src` is for the BridgeStan code, and any folder named after a language is for _accessing_ that code from the language it is named after

WardBrian:
This PR:

1. Closes #26  by renaming `stan/` to `test_models/`
2. Removes some unnecessary portions of the makefile, fixes `make clean`
3. Adds `compile_model` to the Python interface. This takes in the path to a Stan model and compiles it with bridgestan.
4. Adds `StanModel.from_stan_file()` to the Python interface. This is the composition of the above function and the constructor of StanModel.
5. Fixes an issue where the R tests wouldn't properly report failures (It turns out, R only sets exit codes if you ask it to?)

@bob-carpenter For review, I'm mostly wondering if you think I factored the interface in the right way. I considered updating the constructor to take a stan file instead, but I think it's useful the original way we had it set up, hence the class method. Thoughts?


python/bridgestan/model.py
```diff
@@ -0,0 +1,507 @@
...
+            ctypes.POINTER(ctypes.c_double),
+            double_array,
+            double_array,
+        ]
+
+        self._destruct = self.stanlib.destruct
+        self._destruct.restype = ctypes.c_int
+        self._destruct.argtypes = [ctypes.c_void_p]
+
+    @classmethod
```
bob-carpenter, line 168: [comment]
I also didn't know you could declare "static" class functions like this.  The name "classmethod" is terrible, though, as that refers specifically to a non-static class function!

python/example.py
```diff
@@ -26,13 +15,13 @@
...
     for gq in [True, False]:
-        D = model.param_num(include_tp = tp, include_gq = gq)
+        D = model.param_num(include_tp=tp, include_gq=gq)
         print(f"tp = {tp:b}, gq = {gq:b}, number of parameters = {D}\n")
 
 print("PARAMETER NAMES: param_names")
 for tp in [True, False]:
     for gq in [True, False]:
-        names = model.param_names(include_tp = tp, include_gq = gq)
+        names = model.param_names(include_tp=tp, include_gq=gq)
```
bob-carpenter, line 24: [question]
Sorry---forgot to run this all through `black`!  Is there an easy way to automate that with CI?

python/bridgestan/model.py
WardBrian: There is also @staticmethod, which is the same except it doesn't receive `cls` as an input. In this case I suppose the distinction is arbitrary

python/bridgestan/model.py
```diff
@@ -0,0 +1,507 @@
...
+        return cls(str(result), model_data, seed=seed, chain_id=chain_id)
+
+    def __del__(self) -> None:
+        """
+        Destroy the Stan model and free memory.
+        """
+        if hasattr(self, "model_rng") and hasattr(self, "_destruct"):
+            self._destruct(self.model_rng)
+
+    def __repr__(self) -> str:
```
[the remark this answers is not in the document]
WardBrian: `__repr__` controls how the object is echo-d back to you if you use it in a repl or similar, while `__str__` defines what `str(x)` does. The nice thing is if you define repr but not str, str defaults to repr!

python/example.py
WardBrian: The preferred way in the python world is usually a tool called [pre-commit](https://pre-commit.com/). We could set that up here if we wanted, or we could try to do something like that Stan's C++ repositories do and push a commit to your branch to fix it automatically. I don't love those, since you end up with a lot of merge commits and bot commits in the history.

roualdes:
This looks great.  Just want to double check two things.

Am I reading this correctly in that we assume by default that a to be compiled Stan program lives within BridgeStan's test_models folder?

Can we compile models that don't live within BridgeStan's directory structure?  I'm guessing the answer is yes, if the user calls both `set_bridgestan_path()` and `set_cmdstan_path()`.  Just want to double check.  Thanks.

WardBrian:
We can compile a model anywhere, yes. The default value for `BRIDGESTAN_PATH` makes the (perhaps unreasonable) guess that you installed the python package from source, so the bridgestan folder should be a parent of the python package location. This is why I check for the existence of `Makefile`, since that would not appear in, say, Python's `site-packages` directory.

bob-carpenter (approved):

python/bridgestan/util.py
```diff
@@ -0,0 +1,15 @@
...
+    """
+    Raise an error if the specified file is not readable.
+
+    :param f: The file name.
+    :raises FileNotFoundError: If the file is not found.
+    :raises PermissionError: If the file is not readable.
+    """
+    if not os.path.isfile(f):
+        raise FileNotFoundError(f"File '{f}' does not exist")
+    if not os.access(f, os.R_OK):
```
bob-carpenter, line 14: Perfect.  Thanks!

bob-carpenter (approved):

PR #28, opened 2022-09-21
Allow compilation from inside Julia

WardBrian:
This basically does the same thing as #27, but for Julia. Julia's built-in support for running commands is very interesting!

@roualdes - I defined a function called `StanModel_from_stan_file` for this, but that almost certainly is the wrong thing to do. Any suggestions on what that should be called or how it should work? I'm also not very confident in the way I broke this up into files.


julia/docs/src/julia.md
```diff
@@ -28,3 +30,10 @@ param_constrain!
 param_unconstrain!
 param_unconstrain_json!
 ```
+
+## Compilation utilities
+```@docs
+compile_model
+set_bridgestan_path
```
roualdes, line 37:
```suggestion
set_bridgestan_path!
```

roualdes: For `set_cmdstan_path` too, add `!`.

julia/src/compile.jl
```diff
@@ -0,0 +1,97 @@
...
+"""
+    compile_model(stan_file, args=[])
+
+Run BridgeStan’s Makefile on a `.stan` file, creating the `.so` used by StanModel.
+Additional arguments to `make` can be passed as a vector, for example `["STAN_THREADS=true"]`
+enables the model's threading capabilities.
+
+This function assumes that the paths to BridgeStan and CmdStan are both valid.
+These can be set with `set_bridgestan_path()` and `set_cmdstan_path()` if their default
+ values do not match your system configuration.
```
roualdes, line 70: The doc should include what this function returns.

julia/src/model.jl
```diff
@@ -0,0 +1,513 @@
...
+    StanModel(lib, datafile="", seed=204, chain_id=0)
+
+A StanModel instance encapsulates a Stan model instantiated with data.
+
+The constructor a Stan model from the supplied library file path and data file path.
+If seed or chain_id are supplied, these are used to initialize the RNG used by the model.
+
+    StanModel(;stan_file, data="", seed=204, chain_id=0)
+
+Construct a StanModel instance from a `.stan` file, compiling if necessary.
```
roualdes, line 14: I don't see where in the code this sentence fits.

roualdes (commented):
A few follow up questions, but overall this is a great feature to add to our higher-level language APIs.  Thanks.

The addition of `!` to the setters is good.  Will you carry that through the doc, too?

Is there a reason compile.jl has its own module, but model.jl does not?  If you drop the compile.jl module, you won't need (I don't think) 

```
set_cmdstan_path! = Compile.set_cmdstan_path!
set_bridgestan_path! = Compile.set_bridgestan_path!
compile_model = Compile.compile_model
```
in BridgeStan.jl.

Is it recommended somewhere in the Stan-verse to put CmdStan at ~/.cmdstan?  I'm not against this, just want to try to be explicit about the assumptions of this new (and great) compilation of Stan models built into the higher-level language API of BridgeStan.

julia/src/model.jl
WardBrian: This overload is actually defined in BridgeStan.jl (line 31) since it requires both this code and the compile code. I tried documenting it there, which worked for `help` inside Julia, but only documented one of them inside Documenter.jl

WardBrian: I suppose there is no harm duplicating it there?

WardBrian:
> Is there a reason compile.jl has its own module, but model.jl does not

I originally was mirroring python and using global variables which (I think?) needed to live in a module, but this ended up being not very jit friendly since their value would be compiled once and never updated. I'll make the change you suggest.

> Is it recommended somewhere in the Stan-verse to put CmdStan at ~/.cmdstan?

This is the location that cmdstanpy and cmdstanr will install CmdStan to by default. The `$CMDSTAN` environment is also used by those packages, and it is set by `conda`.

julia/src/model.jl
roualdes: I see what's going on.  My bad.  Just missed it / didn't put all the pieces together.  Thanks for helping me see the detail.

julia/src/compile.jl
```diff
@@ -0,0 +1,104 @@
...
+    end
+    if !isfile(joinpath(path, "Makefile"))
+        error(
+            "Makefile does not exist at path! Make sure it was installed correctly.\n$path",
+        )
+    end
+end
+
+"""
+    set_cmdstan_path(path)
```
roualdes, line 37:
```suggestion
    set_cmdstan_path!(path)
```

julia/src/compile.jl
```diff
@@ -0,0 +1,104 @@
...
+function set_cmdstan_path!(path::AbstractString)
+    if !isdir(path)
+        error("Path does not exist!\n$path")
+    end
+    ENV["CMDSTAN"] = path
+end
+
+
+"""
+    set_bridgestan_path(path)
```
roualdes, line 53:
```suggestion
    set_bridgestan_path!(path)
```

julia/src/compile.jl
```diff
@@ -0,0 +1,104 @@
...
+"""
+    compile_model(stan_file, args=[])
+
+Run BridgeStan’s Makefile on a `.stan` file, creating the `.so` used by StanModel and
+return a path to the compiled library. 
+Additional arguments to `make` can be passed as a vector, for example `["STAN_THREADS=true"]`
+enables the model's threading capabilities.
+
+This function assumes that the paths to BridgeStan and CmdStan are both valid.
+These can be set with `set_bridgestan_path()` and `set_cmdstan_path()` if their default
```
roualdes, line 75:
```suggestion
These can be set with `set_bridgestan_path!()` and `set_cmdstan_path!()` if their default
```

roualdes:
Re global variables.  Am I reading this correctly?  Now you aren't using globals, right? ENV variables where necessary and otherwise everything is encapsulated or whatever the phrase is to make globals into local vars?

WardBrian:
That’s correct! My first attempt was close to just a straight copy of the python, but I’ve iterated a couple times

roualdes (approved):

PR #30, opened 2022-09-23
Use a Make rule to build all test models

WardBrian:
This removes `test_models/build_all.sh` and `test_models/build_all.ps1` in favor of simply using a make rule called `test_models` to build all test models. This makes it uniform on all platforms and it allows speedups if you run, e.g. `make test_models -j8`

### DIFF
--- a/.github/workflows/main.yaml
+++ b/.github/workflows/main.yaml
@@ -9,7 +9,7 @@ on:
 
 env:
   CMDSTAN_VERSION: "2.30.1"
-  CACHE_VERSION: 3
+  CACHE_VERSION: 4
 
 jobs:
   build_test_models:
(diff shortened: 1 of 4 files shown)


WardBrian:
@bob-carpenter this changes what we discussed earlier, but hopefully for the better. Now instead of running `./test_models/build_all.sh` the user can just run 

`make test_models`

If they want they can use the usual arguments like `-j8`. It is also worth noting that the Julia tests will assume you compiled with `STAN_THREADS=true`

So, on my machine, the command I run if I want to run the tests is `make STAN_THREADS=true test_models -j8`

roualdes:
Big ups, @WardBrian.  Those makefile shenanigans are clever.  Nicely done.

PR #41, opened 2022-10-04
Replace model_functor with a lambda

### LINKED ISSUE #39: A vector parameter as the first parameter in a Stan model leads to the first two parameters having a zero gradient if make/local contains `STANCFLAGS+=--O1`.
Weirdly specific title, but that's how it appears to be.

Running the following in the folder `bridgestan/julia/mwe` outputs `(-27.5, [0.0, 0.0, -3.0, -4.0, -5.0])` for me:

```
using Pkg;
Pkg.add(url="https://github.com/roualdes/bridgestan.git", subdir="julia")
using BridgeStan
set_bridgestan_path!(pwd() * "/../..")

model_code = "
parameters {
  vector[5] fvb;
}
model {
  fvb ~ normal(0, 1);
}
"
data_string = "{}"


open("model.stan", "w") do f
    write(f, model_code)
end

open("data.json", "w") do f
    write(f, data_string)
end

model = StanModel(stan_file="model.stan", data="data.json")
println(log_density_gradient(model, collect(1.:5)))
```

WardBrian:
Resolves #39.

@SteveBronder and I, after a fair amount of process-of-elimination, got down to determining the error was caused by some detail of the model functor. He's not sure exactly what it is still, but our guess is something had its destructor called too early and that is what was causing the strange behavior and segfaults with `STANCFLAGS=--O1`. 

I've confirmed that the original model reported in #39 now behaves properly, and I locally ran our entire unit test suite with `STANCFLAGS=--O1` without further issues.


WardBrian:
Looking at the code that model_functor was presumably adapted from (`stan/src/stan/model/model_functional.hpp`), the original was assuming the template type `M` would be the model, not a pointer to it. My guess is something gets weird when passing a class member which is a pointer as a const reference to another class, but I'm not nearly well versed enough in C++ memory to say definitively that was the source of the problem.

roualdes (approved):
Looks good to me.  Great work, y'all.  Thanks much.

Issue #40, opened 2022-10-04
add function to return CmdStan and Stan version numbers


bob-carpenter:
We should have a way from within BridgeStan to report which version of CmdStan and which version of Stan it's using.  

For CmdStan, we can get it from the makefile in CmdStan from the home directory:

```
make print-CMDSTAN_VERSION
```

For Stan, it's available as a three C++ constants.

bob-carpenter:
I can add a function to get the Stan version.  Do you think it should return a triple of major/minor/patch or just a number?

WardBrian:
I think it's best to return a string of "major.minor.patch".

Do we want to have a function specifically for the version? I think ultimately the following would be useful:

1. Version (major.minor.patch)
2. What Stan defines were used (see `cmdstan/src/cmdstan/write_stan_flags.hpp` for a similar idea)
3. What Stanc version and flags were used (Available as `model_->model_compile_info()`).


To me it would make sense to have all of this output by the same function called something like `model_info()` rather than splitting it up

WardBrian:
For example, I think it's reasonable for this method to output a string like:

```
Stan version: 2.30.0
Stan C++ Defines:
        STAN_THREADS=false
        STAN_MPI=false
        STAN_OPENCL=false
        STAN_NO_RANGE_CHECKS=false
        STAN_CPP_OPTIMS=false
Stan Compiler Details:
        stanc_version = stanc3 v2.30.1
        stancflags = 
```

I think the code to do this is pretty simple

PR #42, opened 2022-10-05
Add `model_info` function

WardBrian:
This is a proposal for #40. It includes the version, stanc flags, and Stan defines all in one monolithic string. It is based on the `info` subcommand of CmdStan models, with some extra information

This is the most useful version for debugging (e.g., if we want to ask the user to give us some information), but probably the least useful format if we want it to be machine readable, which would be better served with separate functions for all of this.

Example output:

```
Stan version: 2.30.0
Stan C++ Defines:
        STAN_THREADS=true
        STAN_MPI=false
        STAN_OPENCL=false
        STAN_NO_RANGE_CHECKS=false
        STAN_CPP_OPTIMS=false
Stan Compiler Details:
        stanc_version = stanc3 v2.30.1
        stancflags =
```

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -15,6 +15,7 @@ notes.org
 __pycache__/
 .ipynb_checkpoints/
 .pytest_cache/
+.mypy_cache/
 *.egg-info/
 *.pyc
 _build/
--- a/R/R/bridgestan.R
+++ b/R/R/bridgestan.R
@@ -45,6 +45,15 @@ StanModel <- R6::R6Class("StanModel",
       )$name_out
     },
     #' @description
+    #' Get compile information about this Stan model.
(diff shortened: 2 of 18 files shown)


roualdes (approved):
Totally agree, one big string seems fine for now.  If we eventually need specific machine readable compiler info, we can develop it then.

Just to double check: the compiler flags being tested against are ENV vars specified in the GitHub actions, right?

WardBrian:
> Just to double check: the compiler flags being tested against are ENV vars specified in the GitHub actions, right?

For Julia I check the value `STAN_THREADS=true` since the `multi.stan` test actually requires these later on. For R and Python, I just check that the variable name is in the returned string, not a specific value

bob-carpenter:
Fine with me, so I just merged.

Issue #43, opened 2022-10-07
refactor doc into user-facing and dev-facing and by language


bob-carpenter:
The main goal of the refactor is to organize docs into dev-facing and user-facing, make it easy for users and devs to just look at the language they're using or developing, and to put all the documentation source into either top-level or language-specific directories as appropriate.   This can be achieved with the following directory layout.

* `bridgestan/README.md`
    - project overview
    - link to top-level user-facing doc
    - link to top-level dev facing doc

* `bridgestan/CONTRIBUTING.md`
    - link to top-level dev facing doc

* `bridgestan/docs`
    - language-agnostic documentation
    - `docs/users/` for user-facing language-agnostic documentation
    - `docs/devs/` for dev-facing language-agnostic documentation

* `bridgestan/L/docs`
    - top-level doc for language `L` with pointers to dev and user doc
    - `bridgestan/L/docs/users` for user-facing doc for language `L`
    - `bridgestan/L/docs/devs` for dev-facing doc for language `L`

We need to put instructions on how to build the docs locally.  I see all the dependencies listed on the CONTRIBUTING.md page, but I don't see any instructions on how to build.

WardBrian:
Can you clarify what parts of this you intend to appear on the rendered website at https://roualdes.github.io/bridgestan/ and what parts you want to live only on Github?

The website doc has all of its sources in `python/docs` since it uses sphinx, and splitting it across multiple folders will be quite messy.

If we wanted to, we could move the the `python/docs` folder to the top level `docs/`, but it would still be preferable if all the final docs intended to end up online were in the same folder

bob-carpenter:
From @WardBrian on building the doc locally:

Building the docs locally requires installing (at minimum) sphinx, pydata-sphinx-theme, nbsphinx, and myst-parser. These are all available under those names from Pip.

At a maximum, if you want to also build the Julia and C++ doc, you need 'breathe' from pip, doxygen, and Julia 1.8 installed. These will be simply skipped if you don't have them.

Once you have the above installed, the build is done by running `make html` inside the python/docs/ folder. This will create a new sub-directory called python/docs/_build/html, with the files in there.

bob-carpenter:
> Can you clarify what parts of this you intend to appear on the rendered website at https://roualdes.github.io/bridgestan/ and what parts you want to live only on Github?

Sorry that wasn't clear.  I'd be happiest if the .md files are just pointers to the rendered website.

[...]

PR #44, opened 2022-10-07
Move website sources to top-level docs/

WardBrian:
This is the first and simplest part of #43. All it does is move the website sources from `python/docs` to `docs/` and update the associated file paths in config and CI. 

It also adds a top level `make docs` command to build the documentation locally.

It does not make any content changes, I'm leaving the first pass at those to @bob-carpenter


bob-carpenter (approved):
thanks!  I don't have the build stuff installed yet, but I assume you tested that.

Issue #36, opened 2022-10-04
Bridgestan.jl does not update modified Stan models.


nsiccha:
Initializing a Stan model once, then changing the Stan model in-place (i.e. editing the .stan file) and reinitializing the Stan model does not seem to update the loaded library. 

A possible but messy workaround seems to be to give each .so file an additional random string, e.g. in `compile.jl`

```julia
    rstr = randstring(6)
    dummy_file = splitext(absolute_path)[1] * "_$(rstr).stan"
    cp(absolute_path, dummy_file)
    output_file = splitext(absolute_path)[1] * "_$(rstr)_model.so"
```

WardBrian:
I believe POSIX dlopen has this behavior where repeated calls return the same handle due to reference counting. I think you could get around it by manually calling dlclose on the library contained in the model struct, but I’m not sure this is something the bridgestan interface can address

nsiccha:
@WardBrian I think people should know that editing the stan file will silently fail, i.e. this should be put prominently put int othe documentation.

bob-carpenter:
> people should know that editing the stan file will silently fail, i.e. this should be put prominently put int othe documentation.

We can mention that if you change the Stan file you have to recompile.  But this is how users should expect this to work, because it's the way lm and glm work in R (if you edit the file in a fit object, it doesn't auto update), the way CmdStan(R, Py) work (you have to recompile explicitly if you change the Stan program) , and the way C++ itself works.

nsiccha:
@bob-carpenter BridgeStan **does recompile**, IMO making the user think that the recompiled model will also be used, but it will silently use the old model if that has been compiled and loaded before. I believe this is highly unintuitive.

nsiccha:
The following chunk of code in compile.jl fixes the issue (with high probability):
```julia
    rstr = ""
    open(absolute_path) do f
        rstr = bytes2hex(sha256(f))
    end
    dummy_file = splitext(absolute_path)[1] * "_$(rstr).stan"
    if !isfile(dummy_file)
        cp(absolute_path, dummy_file)
    end
    output_file = splitext(absolute_path)[1] * "_$(rstr)_model.so"
```

WardBrian:
Without some good deal of bookkeeping and clean up, the above could would lead to folders filled with "mymodel_{some hash}_model.so" files.

[...]

PR #45, opened 2022-10-17
Add warning in R and Julia for re-loading DSOs

WardBrian:
See discussions in #36 

I can't figure out if Python exposes a comparable way of testing if a DLL is already loaded or not, so for now this is only in the other two interfaces.


roualdes (approved):
LGTM.  Thanks.

PR #49, opened 2022-11-08
Support for Stan 2.31

WardBrian:
Now that the release candidates are out (https://github.com/stan-dev/cmdstan/issues/1126) we can start testing against them and making the required changes for the next version.

We shouldn't merge this until 2.31 is officially released (~18th of November).

Changes:
- JSON now lives in stan repository instead of cmdstan. We can now drop cmdstan dependency, if we want, in a separate PR.
- We can avoid an extra copy being made in the gradient calculations

### DIFF
--- a/.github/workflows/docs.yaml
+++ b/.github/workflows/docs.yaml
@@ -7,7 +7,7 @@ on:
   workflow_dispatch: {}
 
 env:
-  CMDSTAN_VERSION: "2.30.1"
+  CMDSTAN_VERSION: "2.31.0-rc1"
 
 jobs:
   build-docs:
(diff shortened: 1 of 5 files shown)


WardBrian:
I should note that in addition to the above changes, 2.31 will also allow us to have opt-in higher-order autodiff for Hessians, rather than exclusively using finite differences. That will also be a separate PR, based off https://github.com/roualdes/bridgestan/tree/experiment/ad-hessian

WardBrian:
@roualdes You suggested previously you had some thoughts on the code for the copy-free gradient eval, so I want to get your OK here before merging if you get the chance to take a look

roualdes (approved):
It looks like your [61e4509](https://github.com/roualdes/bridgestan/pull/49/commits/61e45090ec4b14a1e715abf8f053dd814f0fa664) caught the minor point I was going to make.

Looks good to me.  Thanks.

PR #51, opened 2022-11-22
Remove CmdStan Dependency

WardBrian:
This removes the dependency on CmdStan. 

BridgeStan now ships with a copy of the Stan repository, and is able to download `stanc` by the same method CmdStan does.

As part of this we need to pick a "blessed" version of Stan to package with BridgeStan. I have picked Stan 2.31.0 for this as it is the latest version, and I imagine we will want to track this as we go along. 

I have also writen up some documentation on how to use custom versions of Stan and Stanc with bridgestan in the installation instructions.

### DIFF
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -48,7 +48,7 @@ If you wish to build the C++ portions of the documentation, you should also have
 
 ## Builds
 
-We use [Gnu make](https://www.gnu.org/software/make/) for builds.  If you have installed [CmdStan](https://mc-stan.org/users/interfaces/cmdstan) as required for the source to work, then you will already have Gnu make and a sufficient C++ compiler.
+We use [Gnu make](https://www.gnu.org/software/make/) for builds.  If you have installed [CmdStan](https://mc-stan.org/users/interfaces/cmdstan), then you will already have Gnu make and a sufficient C++ compiler.
 
 
 ## Language specific practices
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -48,7 +48,7 @@ If you wish to build the C++ portions of the documentation, you should also have
 
 ## Builds
 
-We use [Gnu make](https://www.gnu.org/software/make/) for builds.  If you have installed [CmdStan](https://mc-stan.org/users/interfaces/cmdstan), then you will already have Gnu make and a sufficient C++ compiler.
+We use [Gnu make](https://www.gnu.org/software/make/) for builds.  If you have previously used [CmdStan](https://mc-stan.org/users/interfaces/cmdstan), then you will already have Gnu make and a sufficient C++ compiler.
 
 
 ## Language specific practices
--- a/README.md
+++ b/README.md
@@ -169,7 +169,7 @@ should run without errors.
 
 ```shell
 cd <bridgestan-dir>
-make test_models/multi/multi
+make test_models/multi/multi_model.so
 ```
 
 This will require internet access the first time you run it in order
--- a/README.md
+++ b/README.md
@@ -10,6 +10,7 @@ Stan is a probabilistic programming language for coding statistical
 models.  For an introduction to what can be coded in Stan, see the
 [*Stan User's Guide*](https://mc-stan.org/docs/stan-users-guide/index.html).
 
+BridgeStan is currently shipping with Stan version 2.31.0
 
 More documentation is available at https://roualdes.github.io/bridgestan/
 
@@ -26,11 +27,10 @@ compiler combinations.
 
 ## Installing BridgeStan
 
-### Step 1: Install C++, make, and CmdStan
+### Step 1: Install a C++14 compiler and make
 
 See the section [Prerequisites](#prerequisites) for instructions on
-installing a C++ toolchain and CmdStan.  Please verify CmdStan is
-working before proceeding.
+installing a C++ toolchain.
 
 ### Step 2: Download BridgeStan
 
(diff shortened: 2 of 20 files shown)


WardBrian:
This mirrors the file structure of CmdStan in that `stan` is a subdirectory of the repo, and then we have our own `src` directory.

An alternative I considered was putting `stan` inside `src` and moving the existing source files one directory deeper, something like

```
.
./src/
./src/stan # submodule
./src/bridgestan/
./src/bridgestan/<current contents of src/>
```

Thoughts @bob-carpenter, @roualdes?

bob-carpenter:
> This mirrors the file structure of CmdStan in that stan is a subdirectory of the repo, and then we have our own src directory.

That works for me.

Do you want me to review this?  It seems to still be testing, but I don't see a commit in the last 5 days, so I'm not sure why.

WardBrian:
I pushed a commit about 10 minutes ago which re-triggered the tests. 

I believe this is ready for review. Before we merge I’d like stan-dev/stan to update the `master` branch to the latest release (Nic is going to do this soon). That won’t have any code changes, but it will change the specific SHA hash of the submodule

CONTRIBUTING.md
```diff
@@ -48,7 +48,7 @@ If you wish to build the C++ portions of the documentation, you should also have
 
 ## Builds
 
-We use [Gnu make](https://www.gnu.org/software/make/) for builds.  If you have installed [CmdStan](https://mc-stan.org/users/interfaces/cmdstan) as required for the source to work, then you will already have Gnu make and a sufficient C++ compiler.
+We use [Gnu make](https://www.gnu.org/software/make/) for builds.  If you have installed [CmdStan](https://mc-stan.org/users/interfaces/cmdstan), then you will already have Gnu make and a sufficient C++ compiler.
```
roualdes, line 51: Maybe this is a silly concern, but can we change this sentence to be more clear that a user needs to be able to build, more than just clone and `make stan-update`, a Stan program using CmdStan for BridgeStan to work correctly?

README.md
```diff
@@ -166,48 +152,29 @@ The interfaces may depend on different packages in their respective languages.
...
+Stan program.  We will assume BridgeStan is checked out in `<bridgestan-dir>`.
 
 To verify the compiler chain is installed correctly, the following
 should run without errors.
 
 ```shell
-cd <cmdstan-dir>
-make <bridgestan-dir>/test_models/multi/multi
+cd <bridgestan-dir>
+make test_models/multi/multi
```
roualdes, line 172: `make test_models/multi/multi_model.so` no?

docs/install.rst
```diff
@@ -18,24 +21,38 @@ Installing BridgeStan is as simple as ensuring that the above pre-requisites are
...
+Using custom Stan versions
+--------------------------
+
+If you wish to use BridgeStan for an older released version, all you need to do is
+
+1. Set ``STANC3_VERSION`` in ``make/local`` to your desired version, e.g. ``v2.26.0``
+2. Go into the ``stan`` submodule and run ``git checkout release/VERSION``, e.g. ``release/v2.26.0``
+3. Also in the ``stan`` submodule, run `make math-update`
+
+To return to the version of Stan currently used by BridgeStan, you can run ``make stan-update`` from the top level directory
```
roualdes, line 49: Does it matter if `make stan-update` is run before removing the `STANC3_VERSION` variable?  Based on the itemized list above, I would assume removing `STANC3_VERSION` should happen first, before running `make stan-update`, but that's not how I read this sentence.

julia/src/compile.jl
```diff
@@ -78,11 +45,10 @@ return a path to the compiled library.
 Additional arguments to `make` can be passed as a vector, for example `["STAN_THREADS=true"]`
 enables the model's threading capabilities.
 
-This function assumes that the paths to BridgeStan and CmdStan are both valid.
-These can be set with `set_bridgestan_path!()` and `set_cmdstan_path!()` if their default
-values do not match your system configuration.
+This function assumes that the paths to BridgeStan is valid.
```
roualdes, line 48: path, singular

python/bridgestan/compile.py
```diff
@@ -82,10 +57,8 @@ def compile_model(stan_file: str, args: List[str] = []) -> Path:
     Run BridgeStan's Makefile on a ``.stan`` file, creating the ``.so``
     used by the StanModel class.
 
-    This function assumes that the paths to BridgeStan and CmdStan
-    are both valid. These can be set with :func:`set_bridgestan_path`
-    and :func:`set_cmdstan_path` if their default values do not
-    match your system configuration.
+    This function assumes that the paths to BridgeStan is valid.
```
roualdes, line 60: path, singular

roualdes (approved):
I too like keeping the Stan submodule at the top level, as you've done.

Some minor changes suggested and one, I think, typo (missing _model.so).

Otherwise, looks good to me.  This is a nice simplification.  Thanks.

README.md
WardBrian: Good catch

docs/install.rst
WardBrian: Good point, there are actually a garden of forking paths of orders one could do here. The important thing for stanc is actually running `make clean` or deleting `bin/stanc` to force a re-download, so that should be on the list!

WardBrian:
I can fix those doc things first thing tomorrow and then merge

WardBrian:
Still waiting on stan-dev/stan to update their `master` branch, but I've addressed the above comments

PR #52, opened 2022-11-22
Add the option for autodiff Hessian calculations

WardBrian:
This adds a flag `BRIDGESTAN_AD_HESSIAN` which switches the `log_density_hessian` function to use nested autodiff. By default, central finite differences are still used. 

I added information about this under "Tips" in the README, and a test is added to the Python suite. 

Tests for features like this are a bit annoying, due to the restriction that we can't link DLLs compiled with different source like this into the same Python runtime. I added some pytest configuration which lets us not run this new test by default, and run it with `pytest --run-type=ad_hessian`, which will skip all tests except the one marked with that marker.


bob-carpenter (approved):
looks good.  Thanks for adding the careful doc.  I'm not 100% sure about all the makefile magic, but am just assuming it will work.

bob-carpenter:
I'm going to let @WardBrian merge here, as I'm not 100% sure about the staging of this vs. other things.

cpfiffer:
Oh man I did not see that this was a feature, this is going to make my life so much easier. Thanks for this one!

PR #54, opened 2022-11-28
Prefix any C-exposed symbols with `bs_`

### LINKED ISSUE #53: Prefix C-level functions to avoid collisions
This would only affect users who directly include `bridgestan.h` in a project, but it is useful (and [fairly standard](https://halt.software/libraries-in-c-prefix-or-not/), I believe) to prefix C function names with a common string to avoid namespace issues that can arise in C. So, our C API would have functions like `bs_name`, while Python/Julia/R etc would still just use `name`.

I'm happy to make these changes, the question is just which prefix to use:

- `bs` short, but commonly used for another meaning
- `stan` - still short, `stan_log_density` etc all read naturally, but this isn't an official Stan project
- `bridgestan` - most accurate and unambiguous, but very long

WardBrian:
Closes #53.

This is a breaking change for the C API, but not for any of the higher level APIs.

### DIFF
--- a/julia/src/model.jl
+++ b/julia/src/model.jl
@@ -31,7 +31,9 @@ mutable struct StanModel
         end
 
         if in(abspath(lib), Libc.Libdl.dllist())
-            @warn "Loading a shared object '" * lib * "' which is already loaded.\n" *
+            @warn "Loading a shared object '" *
+                  lib *
+                  "' which is already loaded.\n" *
                   "If the file has changed since the last time it was loaded, this load may not update the library!"
         end
 
--- a/julia/src/model.jl
+++ b/julia/src/model.jl
@@ -44,7 +44,7 @@ mutable struct StanModel
         lib = Libc.Libdl.dlopen(lib)
 
         stanmodel = ccall(
-            Libc.Libdl.dlsym(lib, "construct"),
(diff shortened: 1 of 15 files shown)


julia/src/model.jl
```diff
@@ -57,7 +59,7 @@ mutable struct StanModel
 
         function f(sm)
             ccall(
-                Libc.Libdl.dlsym(sm.lib, "destruct"),
+                Libc.Libdl.dlsym(sm.lib, "bs_destruct"),
```
bob-carpenter, line 62: `construct` -> `bs_construct` (line 47).

bob-carpenter (changes requested):
I only found one place where it looks like the names weren't changed consistently.  I opened all the files and checked everything was changed to `bs_` as appropriate.  Thanks for doing this!

julia/src/model.jl
WardBrian: Good catch, I did a search for the string `sm.lib` but in the constructor itself the library is called only `lib`. I assume the tests would have caught it rather quickly, but I only ran the Python ones locally before pushing

bob-carpenter (approved):
looks good.  and you found a missing one in R that I missed.

PR #58, opened 2022-12-01
Compile methods: split stanc args from make args, add default include path

### LINKED ISSUE #56: Can't use #include in stan files
As above. Trying to compile a model with include directives fails for my (from python, but I don't think it should depend on the interface).

MWE:
```
import bridgestan as bs
import os
bs.set_bridgestan_path('bridgestan')
os.system("cd bridgestan; git log --name-status HEAD^..HEAD; cd ..")

stan_functions_path = "./functions.stan"
stan_model_path = "./model.stan"
data_path = "./data.json"

with open(stan_functions_path, 'w') as fd:
    fd.write("""
functions {
    int test(){return 1;}
}
""")

with open(stan_model_path, 'w') as fd:
    fd.write("""
#include functions.stan
data {

}
parameters {
    real theta;
}
model {
    theta ~ std_normal();
}
""")

with open(data_path, 'w') as fd:
    fd.write("""{}""")


model = bs.StanModel.from_stan_file(stan_model_path, data_path)
```

Fails with output 
```
commit f62bf462175df7411893e45823926a3d04eb9263 (HEAD -> main, origin/main, origin/HEAD)
Author: Brian Ward <bward@flatironinstitute.org>
Date:   Wed Nov 30 12:54:46 2022 -0500

    Fix broken link

M       docs/install.rst

[...]

--- Translating Stan model to C++ code ---
./bin/stanc  --o=/home/niko/github/mwes/bridgestan/model.hpp /home/niko/github/mwes/bridgestan/model.stan
Makefile:49: recipe for target '/home/niko/github/mwes/bridgestan/model.hpp' failed

stderr:
Syntax error in '/home/niko/github/mwes/bridgestan/model.stan', line 2, column 0, include error:
   -------------------------------------------------
     1:  
     2:  #include functions.stan
         ^
     3:  data {
     4:  
   -------------------------------------------------

Could not find include file functions.stan in specified include paths.
make: *** [/home/niko/github/mwes/bridgestan/model.hpp] Error 1


```

WardBrian:
Closes #56.

This changes the Python and Julia `compile_model` methods and their associated constructors of the StanModel objects to allow you to specify arguments for `make` and `stanc3`. Additionally, mirroring cmdstanpy and cmdstanr, we add the current working directory as a default include path.

### DIFF
--- a/docs/languages/julia.md
+++ b/docs/languages/julia.md
@@ -354,15 +354,15 @@ The result is stored in the vector `out`, and a reference is returned. See `para
 
 
 ```julia
-compile_model(stan_file, args=[])
+compile_model(stan_file; stanc_args=[], make_args=[])
 ```
 
-Run BridgeStan’s Makefile on a `.stan` file, creating the `.so` used by StanModel and return a path to the compiled library. Additional arguments to `make` can be passed as a vector, for example `["STAN_THREADS=true"]` enables the model's threading capabilities.
+Run BridgeStan’s Makefile on a `.stan` file, creating the `.so` used by StanModel and return a path to the compiled library. Arguments to `stanc3` can be passed as a vector, for example `["--O1"]` enables level 1 compiler optimizations. Additional arguments to `make` can be passed as a vector, for example `["STAN_THREADS=true"]` enables the model's threading capabilities.
 
 This function assumes that the path to BridgeStan is valid. This can be set with `set_bridgestan_path!()`.
 
 
-<a target='_blank' href='https://github.com/roualdes/bridgestan/blob/main/julia/src/compile.jl#L42-L52' class='documenter-source'>source</a><br>
+<a target='_blank' href='https://github.com/roualdes/bridgestan/blob/main/julia/src/compile.jl#L42-L54' class='documenter-source'>source</a><br>
 
 <a id='BridgeStan.set_bridgestan_path!' href='#BridgeStan.set_bridgestan_path!'>#</a>
 **`BridgeStan.set_bridgestan_path!`** &mdash; *Function*.
--- a/julia/src/BridgeStan.jl
+++ b/julia/src/BridgeStan.jl
@@ -25,13 +25,19 @@ include("model.jl")
 include("compile.jl")
 
 """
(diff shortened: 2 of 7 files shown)


roualdes (approved):
Overall, this looks great.  Just a couple of questions.  The first is not so important, but I ask mostly as a learning opportunity for me (thanks in advance).

1. What is your preference for spaces surrounding equal signs within function definitions/calls?  Over time, I think I've learned from you, no spaces in function definitions and spaces on either side of equals at call sites.  Does julia/src/compile.jl follow this or am I mistaken?

2. A few questions on our expectations of args to be passed to the compile_model functions.

    a. By my reading, make and stanc args specified in compile_model function calls will take precedence over the same args specified in a Makefile, is this correct?  If I'm right, this is reasonable to me, but I think we should add it to the doc.  If I'm wrong, will you add to the doc for however it does work?

    b. Can multiple instances of `--include-paths=.` be used, say `--include-paths=. --include-paths=/home/stan_fns`?  Maybe this is more of a stanc question.

Much thanks for this.

WardBrian:
> What is your preference for spaces surrounding equal signs within function definitions/calls? Over time, I think I've learned from you, no spaces in function definitions and spaces on either side of equals at call sites. Does julia/src/compile.jl follow this or am I mistaken?

My preference is simply "whatever a good formatting tool outputs". Thus far I've been using [`JuliaFormatter`](https://github.com/domluna/JuliaFormatter.jl), though I'll admit to not being too happy with how slow it is/how poorly it integrates into my editor

> By my reading, make and stanc args specified in compile_model function calls will take precedence over the same args specified in a Makefile, is this correct? 

Yes. Things on the command line will take precedence over `make/local`, which itself takes precedence over environment variables. Good idea to mention this in the documentation.

> Can multiple instances of --include-paths=. be used?

Huh, I thought this was the case but it seems not to be. All but the last is ignored, if you want to put multiple you need to pass them in the same invocation, separated by commas like `--include-paths=.,/home/stan_fns`.

So we should definitely move the default of `.` up to the start, so that later invocations will override it rather than the other way around

roualdes (approved):
Great.  Thanks so much!

PR #61, opened 2022-12-07
[Release] Version bumps and doc for 1.0 release

### LINKED ISSUE #19: Any remaining hurdles to a release?
Continuing the conversation from #18 about when we think BridgeStan is ready for an official release and a version number.

Quoting @WardBrian.

> After this PR I personally consider this repo to be in a "releasable" state, at least for a 0.1 release if not a 1.0

Seconded.

> There are a few things remaining that I have on the list of possible features going forward, such as improved error messages (optional), and utilities in the interfaces for running compilation for you (would be very useful)

The latter would be a great feature, and improved error messages wouldn't hurt.  Both though are features, in my mind, beyond the current usable state of BridgeStan.  

Hence, I'm in favor of a 0.1 or 1.0 version number.

Any other thoughts, opinions, concerns?

WardBrian:
This closes #19 and should be the last PR merged before BridgeStan 1.0.0 is released. 

A draft release is already visible (if you have write access to the repo) on the releases page https://github.com/roualdes/bridgestan/releases.

After this is merged I will add the current state of `main` as a `.tar.gz` file to that release and then we can publish it.

Changes:
- Set versions in each interface to 1.0
- Added text to the getting started page about downloading from the Releases page 
- Added a `version.hpp` based off of Stan's `version.hpp` and included the current BridgeStan version in `model_info`.

### DIFF
--- a/docs/getting-started.rst
+++ b/docs/getting-started.rst
@@ -20,8 +20,18 @@ Downloading BridgeStan
 Installing BridgeStan is as simple as ensuring that the above requirements are installed and then downloading
 the source repository.
 
-Installing with ``git``
-_______________________
+Downloading a released archive
+______________________________
+
+Downloads of a complete copy of the source code and interfaces are available
+on `our GitHub releases page <https://github.com/roualdes/bridgestan/releases>`__.
+
+To use these, simply download the file associated with the version you wish to use,
+and unzip its contents into the folder you would like BridgeStan to be in.
+
+
+Installing the latest version with ``git``
+__________________________________________
 
 If you have ``git`` installed, you may download BridgeStan by navigating to the folder you'd like
 BridgeStan to be in and running
--- a/src/version.hpp
+++ b/src/version.hpp
@@ -1,0 +1,31 @@
+#ifndef BRIDGESTAN_VERSION_HPP
+#define BRIDGESTAN_VERSION_HPP
+
+#include <string>
+
+#ifndef BRIDGESTAN_STRING_EXPAND
+#define BRIDGESTAN_STRING_EXPAND(s) #s
+#endif
+
+#ifndef BRIDGESTAN_STRING
+#define BRIDGESTAN_STRING(s) BRIDGESTAN_STRING_EXPAND(s)
+#endif
+
+#define BRIDGESTAN_MAJOR 1
+#define BRIDGESTAN_MINOR 0
+#define BRIDGESTAN_PATCH 0
+
+namespace bridgestan {
+
+/** Major version number for Stan package. */
+const std::string MAJOR_VERSION = BRIDGESTAN_STRING(BRIDGESTAN_MAJOR);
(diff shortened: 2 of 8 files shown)


docs/getting-started.rst
```diff
@@ -20,8 +20,18 @@ Downloading BridgeStan
 Installing BridgeStan is as simple as ensuring that the above requirements are installed and then downloading
 the source repository.
 
-Installing with ``git``
-_______________________
+Downloading a released archive
```
bob-carpenter, line 23: Thanks much for adding this and breaking it out.

src/version.hpp
```diff
@@ -0,0 +1,31 @@
...
+#define BRIDGESTAN_VERSION_HPP
+
+#include <string>
+
+#ifndef BRIDGESTAN_STRING_EXPAND
+#define BRIDGESTAN_STRING_EXPAND(s) #s
+#endif
+
+#ifndef BRIDGESTAN_STRING
+#define BRIDGESTAN_STRING(s) BRIDGESTAN_STRING_EXPAND(s)
```
bob-carpenter, line 11: [question]

Why are there two functions here when the second just looks like it delegates to the first?  

Is all this complexity being introduced so we can use both `BRIDGESTAN_MAJOR` as a number and use `bridgestan::MAJOR_VERSION` as a string?   But if they're always going to be numbers, why do we need the strings?  I think the strings are useful for values like "alpha" or "beta".

bob-carpenter (approved):
Looks great.  I had a lingering question, but it shouldn't hold this up.

src/version.hpp
WardBrian: I copied what was done in [`stan/version.hpp`](https://github.com/stan-dev/stan/blob/develop/src/stan/version.hpp). The macros in that file look to be quite old, so I'm not 100% sure what the second level of indirection does for us.

Yes, the idea is that `BRIDGESTAN_MAJOR` as a preprocessor define could be used at compile time, if someone for example was writing code which did something different depending on the version of bridgestan used. The `bridgestan::MAJOR_VERSION` is the equivalent but namespaced and more runtime friendly, though I do not know why it is a string specifically.

bob-carpenter:
These were strings to allow us to have versions like "alpha".  It's OK to leave this as is---we don't need to sort out why Stan has an extra level of indirection.

WardBrian:
@bob-carpenter @roualdes I've pressed "publish" on the releases page. 

We should probably do a new forum post and a post on Andrew's blog soon. I'll let @roualdes have the first claim to write either, but I'm happy to do a forum post if he passes on it

roualdes:
Thanks to you both for your hard work to get us to the 1.0 release.  

@WardBrian, if you're so willing, please do write up a new forum and blog post.  If not, hopefully Bob is willing.  I'm slammed at the end of the semester.

PR #63, opened 2023-01-09
fix links to python & julia examples in readme

mike-lawrence:


bob-carpenter (approved):
thanks for the transposition!

bob-carpenter:
This is OK to merge when tests pass.

PR #65, opened 2023-01-13
Add Julia doc to emphasize compile_model args in StanModel; re issue #64

roualdes:
Re Issue #64, duplicate some doc to emphasize that `StanModel` can accept and pass to `compile_model` the arguments to tweak compilation, `make_args`, and Stanc transpilation `stanc_args`.

PR #66, opened 2023-01-14
change JSON object to JSON string literal for all BridgeStan owned files

roualdes:
There are some occurrences of JSON object in files not directly owned by BridgeStan, so I left those out.

### DIFF
--- a/R/R/bridgestan.R
+++ b/R/R/bridgestan.R
@@ -10,7 +10,7 @@ StanModel <- R6::R6Class("StanModel",
     #' @description
     #' Create a Stan Model instace.
     #' @param lib A path to a compiled BridgeStan Shared Object file.
-    #' @param data Either a string representation of a JSON object or a path to a data file in JSON format ending in ".json".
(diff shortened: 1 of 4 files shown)


bob-carpenter (changes requested):
I'm confused because the other PR said there were 17 instances:

https://github.com/roualdes/bridgestan/pull/65#discussion_r1070426027

but this PR is only addressing a handful.  If this is the full list, I'm happy to merge.

roualdes:
Right, I should have been more clear in this PR.  Effectively, the number 17 was wrong.  Some of the 17 occurrences found lived within the Stan submodule, hence in files not directly owned by BridgeStan.  I only changed the occurrences of JSON object in files that are directly owned by BridgeStan, a subset of the 17.    Does that make sense?

bob-carpenter:
Yes, thanks.  We'll have to clean up in the rest of Stan, too.  As a language dev, I'm particularly sensitive to the distinction between language and reference.

bob-carpenter (approved):

PR #72, opened 2023-01-27
Add compatibility with Julia LTS

sethaxen:
Currently the Julia interface only supports Julia versions 1.8 and newer because of its use of `const` annotation for 3 fields of `StanModel`. This Julia bound will propagate to all dependants of this package at Pkg resolve time, so e.g. if StanSample adds BridgeStan as a required dependency, they will need to drop support for Julia v1.7 and earlier.

Currently the LTS is v1.6, and most Julia packages support at least the LTS unless they have critical functionality that requires newer Julia features. In my opinion the utility of making these 3 fields constant is not worth the limitations it imposes on dependents, but in this PR, I demonstrate how a macro can be used to conditionally annotate those fields only on v1.8 and newer. Locally, the test suite passes now for at least v1.6.

Thanks to @MasonProtter for writing the macro!


WardBrian:
Merging #70 caused a merge conflict here, but if you can fix this it looks phenomenal!

sethaxen:
Done! I also checked whether older versions of Julia could be supported, but no, `Libc.Libdl` was introduced on v1.6. This is fine; many packages are bumping their lower bounds to v1.6, so I don't think this will block others from depending on BridgeStan.

WardBrian (approved):
Thanks!

I'm going to look into Python packaging options soon and hopefully publish/register this week

PR #73, opened 2023-01-31
Add unified release action

WardBrian:
This builds on @sethaxen's work in #70 to create a unified release action for the repo. It also updates a few documentation pages in anticipation of the release.

The steps the Action takes are roughly:

1. Replace the version numbers in all interfaces with the new version
2. Build the `.tar.gz` file we attach to a release
3. Update the Julia `Artifacts.toml` file
4. Push a commit to the repo with the above changes
5. Create a tag and release
6. Push to PyPI (@roualdes this will require us adding a Secret to the repository for a PyPI token once)
7. Comment on the commit from `4` with a `@JuliaRegistrator register`  (@roualdes this will require following the pre-registration instructions here https://github.com/JuliaRegistries/Registrator.jl once)

If any step up to and including 5 fails, you can re-run the script without any problem. If any step _after_ 5 fails, you need to check a box on the Workflow page while running which will skip re-creating the release, and will just upload to PyPI/JuliaRegistries

After the automated release we can manually edit the releases page to include the change log.

After this is merged I suggest we more or less immediately run it to release `1.0.1` which will start the Julia Registration process. We can also discuss if we want to eventually make the Python interface able to download the C++ source, like the Julia interface now can, but I decided to leave that for later.

### DIFF
--- a/.github/workflows/release.yaml
+++ b/.github/workflows/release.yaml
@@ -98,4 +98,4 @@ jobs:
         with:
           sha: ${{ steps.commit.outputs.sha }}
           body: |
-            @JuliaRegistrator register
+            @JuliaRegistrator register subdir=julia
--- a/.github/workflows/release.yaml
+++ b/.github/workflows/release.yaml
@@ -1,0 +1,101 @@
+name: Release BridgeStan
+
+on:
+  workflow_dispatch:
+    inputs:
+      new_version:
+        description: 'New version, for example: 1.1.0'
+        required: true
+      is_rerun:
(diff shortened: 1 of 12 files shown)


roualdes (approved):
This looks good to me. Thanks for giving a go at setting this up.

+1 to immediately release 1.0.1 to get the Julia Registration process kickstarted.

WardBrian:
@roualdes I just wrote to you about some config details for this that need to be done in repository settings before this would work fully

PR #75, opened 2023-02-07
Do not create ChainableStacks except when threading

WardBrian:
From the [Math Library documentation](https://mc-stan.org/math/structstan_1_1math_1_1_autodiff_stack_singleton.html):

> Whenever STAN_THREADS is defined then all created child threads must instantiate a [AutodiffStackSingleton](https://mc-stan.org/math/structstan_1_1math_1_1_autodiff_stack_singleton.html) object within the child thread before accessing the AD system in order to initialize the TLS AD tape within the child thread.

We have been creating a `thread_local` AD stack even when `STAN_THREADS` is not defined. This is unnecessary, and probably incurs some extra overhead.


roualdes (approved):
LGTM.  Appreciate the thoughts/efforts to make things as fast as possible.

PR #78, opened 2023-02-23
example.R: add missing instructions

jgabry:
The instructions at https://roualdes.github.io/bridgestan/languages/r.html#example-program are incomplete. This adds the missing instructions necessary to run the example without errors. I added them as comments in example.R but it could also just go in the text in the markdown file.


WardBrian (approved):
Thanks!

PR #81, opened 2023-03-10
Make sure -fPIC is set globally so Sundials can link

WardBrian:
Reported by @modichirag, models which use ODE integrators from Sundials currently fail during the link step. This is because the static libraries in the Math library were not built with `-fPIC`.

This PR:
1. Globally sets -fPIC in CXXFLAGS
2. Cleans up the makefile a bit to remove unnecessary variables
3. Adds a test model from the Stan User's Guide which will be built in CI

### DIFF
--- a/.github/workflows/main.yaml
+++ b/.github/workflows/main.yaml
@@ -8,7 +8,7 @@ on:
   workflow_dispatch: {}
 
 env:
-  CACHE_VERSION: 1
+  CACHE_VERSION: 2
 
 jobs:
   build_test_models:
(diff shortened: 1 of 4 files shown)


roualdes (approved):
Overall, this looks good.  And cleaning up useless make variables is always helpful.

Let me just check one piece for my understanding.

This PR sets global `CXXFLAGS` which are not implicitly loaded with `$(CC)`, so technically this PR builds different C example models, since c-example/Makefile no longer has `-fPIC`.  However, that doesn't matter because the only necessary shared libraries come from the BrigdeStan source .cpp files, for which `CXXFLAGS` *is implicit*.

WardBrian:
The c example should be the same, since it’s an executable we don’t need the code to be position independent. The change to use CC is just to avoid hard coding gcc, but those specific invocations never used -fPIC

But generally your understanding sounds correct. Defining fPIC as part of CXXFLAGS makes it apply for all the places we already had it, plus it also now affects the sundials library we depend on (through the Math library)

PR #86, opened 2023-03-21
adding CITATION file

bob-carpenter:
I added @roualdes, @sethaxen, and @WardBrian as authors.  We could add @mike-lawrence, too---he added some doc.  

The `CITATION.cff` file should produce a pull-down with citation info on the site.  See: https://docs.github.com/en/repositories/managing-your-repositorys-settings-and-features/customizing-your-repository/about-citation-files

### DIFF
--- a/CITATION.cff
+++ b/CITATION.cff
@@ -1,0 +1,16 @@
+cff-version: 1.2.0
+message: "If you use this software, please cite it as below."
+authors:
+- family-names: "Roualdes"
+  given-names: "Edward"
+- family-names: "Ward"
+  given-names: "Brian"
+- family-names: "Axen"
+  given-names: "Seth"
+- family-names: "Carpenter"
+  given-names: "Bob"
+  orcid: "https://orcid.org/0000-0002-2433-9688"
+title: "BridgeStan: Efficient in-memory access to Stan programs through Python, Julia, and R"
+version: 1.0.1
+date-released: 2023-02-01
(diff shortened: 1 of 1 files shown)


mike-lawrence:
Oh, I only fixed transposed urls in the readme, no need to add me for that!

WardBrian:
We will need an Zenodo (or equivalent) DOI eventually if we get accepted at JOSS:
https://joss.readthedocs.io/en/latest/submitting.html#the-review-process

It might be worth @roualdes getting everything set up to do that now (I'm assuming he needs to as the repo owner, I've never gone through the process myself). 
Those DOIs are specific to a version, so we'd need another one for JOSS if they requested any changes, but it wouldn't hurt to get one going for now?

WardBrian:
It seems like @roualdes can log into Zenodo with Github and then set it up such that this repo will have a DOI for every automatically published release. 

It seems very easy, but it's another one of those things that requires repository admin rights (I can log into Zenodo but then I only see my own repos and things on stan-dev)

roualdes:
I just tried to set up Zenodo such that this repo will have a DOI for every automatically published release.  Maybe we need a new release to kick-start things?

WardBrian:
Since we just added the ability for the python interface to download the source it would be appropriate to do a release soon anyway, I can try it tomorrow?

roualdes:
Sounds good to me.

PR #87, opened 2023-03-22
Add README badges for docs, DOI, build status

WardBrian:
This adds the following badges to the README (discussed https://github.com/roualdes/bridgestan/pull/86#issuecomment-1479613076):

[DOCS](https://roualdes.github.io/bridgestan/) [DOI](https://doi.org/10.5281/zenodo.7760173) [CI](https://github.com/roualdes/bridgestan/actions/workflows/main.yaml)

If we wanted to really go maximalist there are also commonly used badges for licenses, pypi downloads, etc. I don't believe Julia has any for their registry.


roualdes:
I'm not against the maximalist approach, but I think the badges listed in this PR are a good compromise between none and all.

bob-carpenter (approved):
Looks good.  Thanks!

PR #88, opened 2023-03-23
Add a rust interface

aseyboldt:
First, thank you for this library! I've been trying to get better stan integration into nutpie, and this is exactly what I've been looking for!

This (draft) PR adds a rust interface. Is this something you would want to have in bridgestan itself? If not I'll just put it in nutpie, but if other people would like to use this, I thought it would be better to have it here.

The biggest missing piece of this is that I didn't implement any tests yet, I thought I'd ask first if you want this before I spend all the time fighting the compilation in the github actions...

I've got a couple of questions about the C API though:

- I'm not sure about the thread safety of the C functions. I guess if `STAN_THREADS` in not set, then nothing is thread safe. But what are the semantics with `STAN_THREADS=1`? I think there are some thread local variables that need to be initialized somewhere for the stack allocator in stan, right? I'm guessing it is unsafe to call `bs_log_density` from different threads at the same time, but is it also unsafe to call `bs_construct` in one thread, and then `bs_log_density` on the returned pointer in a different thread (ie "is `bs_model_rng` `Send`")? I think it would also be helpful if the value of `STAN_THREAD` was exposed somewhere (eg `bs_multithreading_enabled() -> bool` or so). Since rust wants to know these things at compile time, I'm assuming here right now that `STAN_THREADS=1`, but I can not add an assert to make sure this is actually the case at runtime.
- `bs_construct`: `data_file` is declared to be non-const, I guess that's not actually necessary, is it?
- Same for all the `bs_model_rng` pointer arguments. Probably not all of those have to be mutable? (if they are it makes writing rust programs that use this more tricky sometimes).
- Shouldn't the `theta` parameter of `log_density`, `log_density_gradient` and `log_density_hessian` be called `theta_unc`? Independent of the name, this is on unconstrained space, is it?
- It would be really helpful if there was a way to get the error messages through the API. If this is used in a jupyter notebook for instance, it's not nice to users to tell them that the error messages are somewhere on the console where jupyter was started. If there was an additional field ` *char last_error_msg` in `bs_model_rng`, couldn't there be a `bs_last_error_msg` function then that returns that message?


WardBrian:
This is great, I have a branch which sketched out something very similar: https://github.com/WardBrian/bridgestan/tree/examples/rust-example

It even has an example program using nuts-rs. I went a bit more on the automated side with bindgen and a mostly-working build.rs

I struggled with Send/Sync question. I think calling from multiple threads is fine (and my testing suggests this is true), but I believe free needs to be called on the same thread as the original malloc. 

The `compile_info` function will contain if stan threads was used at compile time, granted as a string. 

We probably don’t have const correctness completely right. Definitely the data string could be (should be) const. Log density family of functions won’t modify the model either, but param_constrain could as it will advance the RNG.


Theta_unc would be a better name for those, yes. 

Getting the message back in a thread safe way is very tricky. The “last_error” approach would most likely cover a lot of cases, but obviously fails if multiple errors are occurring simultaneously 


I’m not sure whether or not this is something we’d merge into the main branch (I am only just learning rust, so we’d not really have a good maintainer for it), but I would love to discuss the implementation more!

roualdes:
> Is this something you would want to have in bridgestan itself?

Speaking only for myself, absolutely.  Though Brian does bring up a good point about our lack of a maintainer knowledgeable in Rust.  I know little about Rust and almost none of the language.

`STAN_THREADS` is meant to be boolean, not integer.  I don't know what happens if one sets `STAN_THREADS` to an integer.  

I've experimented with calling `bs_log_density_gradient()` from up to 12 threads in a Julia program and everything has worked out well, as far as I can tell.  But I'm not sure if this addresses any of your concerns re "I'm guessing it is unsafe to call bs_log_density from different threads at the same time, but is it also unsafe to call bs_construct in one thread, and then bs_log_density on the returned pointer in a different thread"

aseyboldt:
@WardBrian Awesome, should have checked the branches sooner! :grinning: 

I guess the biggest difference (apart from the fact that your branch is much cleaner overall) is that this is opening the dynamic library at runtime using `libloading`. I don't want to assume that users of nutpie have rust installed, and I'd like to avoid recompiling the sampler itself for every model anyway, so I think that would be better at least for my use case. It makes it possible to rely on the compilation code in python (or something else I guess), and only pass the compiled library to the already compiled rust code. That still works with bindgen (I have a call to that in the Makefile as well), see [here](https://docs.rs/bindgen/latest/bindgen/struct.Builder.html#method.dynamic_library_name)

About the thread safety:
I guess the stack allocator is initialized each time the log_density functions are called, so those shouldn't be an issue after all...
https://github.com/roualdes/bridgestan/blob/main/src/model_rng.cpp#L297

So maybe we can assume that `bs_model_rng` is Sync, but not Send (due to the free in the same thread).

> Getting the message back in a thread safe way is very tricky. The “last_error” approach would most likely cover a lot of cases, but obviously fails if multiple errors are occurring simultaneously

True, I just assumed that `bs_model_rng` is not thread safe, so there'd have to be different bs_model_rng instances for each thread anyway. I guess a solution that does not introduce thread safety issues would be to add a `char** error_msg` argument to all functions that could fail? That would be a breaking change, not sure how stable you consider the API to be at the moment. For what it's worth, I think error messages are pretty great though. :-)

And on the topic of errors: Is there a distinction in stan for fatal and non-fatal errors? Ie, something went wrong in the logp evaluation, but we just want to treat that as a divergence and continue to sample, or something really failed and we just want to stop everything?

I see the point about maintenance. If you end up adding a rust backend and something breaks you can definitely ping me if you like, and I'd try to make some time for it, but I can't make any promises. My main focus really is around the PyMC ecosystem.

WardBrian:
My knowledge of libloading only dates back to about a week ago, so when I first tried a rust wrapper I just did a port of how our C example works. I think libloading is totally the way to go for something like nutpie, but I could see people wanting it either way if they were developing in Rust

Correct, the static thread local there ensures there is always a autodiff stack available for each thread (the Stan Math library implements this as a thread-local singleton, so actually only one of those calls will be meaningful, we just need to make sure it exists before we do anything the first time). Instantiating a model once per thread is also totally valid, but for models with large data that will end up hitting memory pretty hard

Avoiding a “you need to call this later to free the memory of an error” is why we made our current choices - avoiding double pointers in general made it much easier to use in Python etc. But they’re definitely not ideal and I would be happy to entertain other options. Especially as we’re imagining more and more end users (as opposed to algorithm developers) using bridgestan) this gets more and more important to get right. 

I believe in Stan we treat all errors as divergences, but I think there must be some escape hatch where if too many happen then the algorithm aborts early. @bob-carpenter will know better

WardBrian:
I've updated my branch to also use libloading, which is much nicer than what I had before!

aseyboldt:
Closing in favor of #102

PR #89, opened 2023-03-24
Fix: Const correctness in C api

WardBrian:
As pointed out by @aseyboldt in #88, our C headers could have many more things marked as `const` than they currently did. This fixes that and cleans up some argument names, also per his suggestion.

### DIFF
--- a/src/bridgestan.cpp
+++ b/src/bridgestan.cpp
@@ -2,7 +2,7 @@
 #include "model_rng.cpp"
 #include "bridgestanR.cpp"
 
-bs_model_rng* bs_construct(char* data_file, unsigned int seed,
+bs_model_rng* bs_construct(const char* data_file, unsigned int seed,
                            unsigned int chain_id) {
   try {
     return new bs_model_rng(data_file, seed, chain_id);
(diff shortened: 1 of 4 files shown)


WardBrian:
I tested locally to confirm we _cannot_ make `param_constrain` const, and with everything passing in CI I think this is good enough

bob-carpenter:
We could get more extreme on this given that C++ lets you cast away `const`.  The biggest problem is in our constructed model class.  Despite being const, I don't believe the log_prob functions are not marked as const.  What we can do in our interface is take in something that's marked `const` and then cast away the `const` modifier, which we know will be safe.

WardBrian:
I think all of the truly const parts of the model class are marked as such. `write_array` requires a non-const RNG object, which is what prevents us from making it const in `param_constrain`

bob-carpenter:
Nice.  I hadn't realized that has been fixed.  And it looks like we can't make the params arg to `log_prob` const because it'll be modified if it's an autodiff variable (though I'm wondering how we can declare all of our math library as const because those also get modified due to autodiff).

Issue #90, opened 2023-03-24
Improved error handling in the C API


aseyboldt:
There was some discussion about error handling in #88, but I thought a separate issue would be a better place to discuss this.

Currently, the error messages of the C++ exception are printed to stderr, which works fine, but has the disadvantage that many users might not know to look there, and depending on the setup it might not even be easy to find it (for example in a jupyter notebook that is started from jupyterhub or from a system service). Also, if several models are failing at the same time, a user would have to figure out which error belongs to which model manually.

The two (I think) most important kinds of failures would be
1. Errors during model creation (ie `bs_construct`). Users will I think provide invalid data fairly often, and the error message should contain information about what part of the data was invalid.
2. Errors during `bs_log_density_gradient` (or for different sampling / VI algorithms the other log_density functions): Error messages often contain information about why there was a divergence during sampling, and sampling algorithms might want to collect this information to help model debugging.

I can see a couple of different options of how this could be achieved, my favorites would probably be those:

1. Store the error information in the `bs_model_rng` struct as an extra field, and provide an accessor function for it. This unfortunately makes the `bs_log_density_gradient` function family no longer thread safe. We can work around that issue however by creating new `bs_model_rng` structs for each thread. With the current API this would require loading the data several times however. This again can be avoided by splitting `bs_model_rng` into two separate types: One for the dataset and the potential error message from model creating, and a separate context for rng and error messages during sampling.
  This would then look like this:
  ```C
  bs_model *model = bs_model_construct(data)
  if (model == 0) {
      // A failure here indicates that we couldn't even allocate the error msg,
      // which would usually not happen...
      return -1;
  }

  // Callers have to check that there wasn't an error explicitly again!
  const char *msg = bs_model_error_msg(model);
  if (msg != 0) {
      // Data was invalid...
      // Do something with message
      bs_model_free(model);
  }

  // In each thread do the following:
  {
      bs_ctx *ctx = bs_ctx_construct(model, seed, seed_idx);
      if (ctx == 0)
          stop();
      }

      // Call log_density and other functions in the sampler or...
      for (int i = 0; i < 10; i++) {
          int rc = bs_log_density(ctx, ...);
          if (rc != 0) {
              char *msg = bs_ctx_error_msg(ctx);
              // do something with message
          }
      }

      bs_ctx_free(ctx);  // Frees the msg field if necessary
  }

  // And finally a single call to free the memory in the
  // original thread
  bs_model_free(model);  # Frees the msg field if necessary
[...]

PR #91, opened 2023-03-27
Breaking change: Propagate error messages through C API

WardBrian:
This is an implementation of "option 2" described in #90.

Every function which can error accepts a `char** err_msg` out param. When an error occurs, 
1. If `err_msg` is `nullptr`, the error message is swallowed
2. Otherwise, it is set to a newly allocated string describing the error. 

Users of the API are required to free this message. This is done with a new function `bs_free_error_msg`, which must be called if the return code was nonzero and `err_msg` was not passed `nullptr`.

_R note:_ R handles strings oddly. The R documentation says (in very few words) that "character vectors" are copied on call and on return for FFI calls. As a result, the R interface has _two_ out params: the error message, and a `void**` pointer which points to the same memory and allows it to be freed.

### DIFF
--- a/julia/src/model.jl
+++ b/julia/src/model.jl
@@ -586,7 +586,7 @@ function log_density_hessian(
 end
 
 """
-    handle_error(sm::StanModel, err::Ref{Cstring}, method::String)
+    handle_error(lib::Ptr{Nothing}, err::Ref{Cstring}, method::String)
 
 Retrieves the error message allocated in C++ and frees it before returning a copy.
 """


WardBrian:
Note, by "users of the API", I mean the C-api, which means the "users" are primarily ourselves in the client libraries. 

The breaking change is also only at this C-level, code which uses our Python/Julia/R interfaces looks exactly the same as before. Users of Bridgestan in Python/Julia/R do not see any change following this PR besides better error messages. They do not need to worry about freeing any memory themselves, the library code handles that when necessary.

WardBrian:
So, if we wanted to release this it should probably be a 2.x, even though it only affects users of the underlying C API. We may want to make more breaking changes (e.g. #92) but assuming this is the only one for now I think the release notes should be something like:

---

## Breaking changes

There are no breaking changes in the Julia, Python, or R interfaces this version. 

The BridgeStan C API has changed to have an additional output parameter for functions which can have errors.  This argument is a pointer to a string (`char **`) which is set when an error is encountered. If it is set, the user must call the (_new_) `bs_free_error_msg` function on the string it points to. Passing in `nullptr` (`0`) for this argument disables the ability to get error messages, but also removes the obligation of the caller to free memory.

See the updated C example program for usage: LINK

roualdes:
I think I'd prefer to wrap this PR, the interface level follow ups to this PR, and #92 into 2.x release.  

My understanding is that, although this PR itself doesn't have any breaking changes in the interfaces, follow ups from this PR at the interface level will have breaking changes in the interfaces, e.g. user specified flags to dictate how printing should happen.  Right?

WardBrian:
The only breaking change at the interface level I've currently been looking at is changes to the signature of `param_constrain` which would require a seed or RNG object in the high-level interfaces (#92). 

Everything else (e.g. whatever we figure out for printing, if anything, #93) should be backwards-compatible. 

This PR already does all the changes to the interfaces needed for better error handling (#90)

WardBrian:
I believe this is now ready for review. The other breaking change (#98) is now up, but in order to avoid having many merge conflicts I based it off this branch and it will require this one to be merged first

WardBrian:
I do not know what a callback approach would look like for error messages. Could you elaborate @bob-carpenter?

WardBrian:
I think we would have two difficulties with that approach.

First, as far as I am aware, it is not possible to pass a callback to C from R without using R-specific headers.

Secondly, the handling on the higher level side would be sort of strange. Unlike with printing as we're discussing in #93, there isn't an immediate action we want to take in the higher-level-language (e.g., in Python, the callback function itself cannot raise an exception, since the Python runtime will just ignore it here). 
The goal is to raise a Python/Julia/R exception with the same string, not just print it and then throw a generic exception like we do now, so the callback would end up looking like (in python, I believe something similar is possible in Julia but I have not checked)

```python
@ctypes.CFUNCTYPE(None, ctypes.c_char_p, ctypes.c_void_pt)
def error_callback(s) :
    # store error message somewhere, most likely a global?
    global LAST_ERROR
    LAST_ERROR = ctypes.string_at(s).decode("utf-8")
```

and then the call sites would need to check wherever we stashed it to format the exception, which now has the classic problems where concurrent calls will overwrite each other's errors.


I don't think this ends up being simpler or cleaner than requiring the memory to be freed.

WardBrian:
@roualdes thoughts on this set of changes?

julia/src/model.jl
```diff
@@ -77,8 +80,9 @@ mutable struct StanModel
             ccall(
                 Libc.Libdl.dlsym(sm.lib, "bs_destruct"),
                 UInt32,
-                (Ptr{StanModelStruct},),
+                (Ptr{StanModelStruct}, Ref{Cstring}),
                 sm.stanmodel,
+                C_NULL,
```
roualdes, line 85: Why not "catch" and handle the error message similarly here, with `_handle_error(lib, err, "bs_descruct")`?

R/R/bridgestan.R
```diff
@@ -249,3 +263,13 @@ StanModel <- R6::R6Class("StanModel",
   ),
   cloneable=FALSE
 )
+#' Get and free the error message stored at the C++ pointer
+#' @keywords internal
+handle_error <- function(function_name, err_msg, err_ptr, lib_name) {
```
roualdes, line 268: Is it worth arranging the arguments in the same(-ish) order as in Julia and Python?  The other languages have `function_name` last and something-like `lib_name` first.

roualdes (approved):
This looks good to me.  So long as all are agreeable to the strategy taken here, then I say we move ahead.

julia/src/model.jl
WardBrian: Good question! Two reasons:

1. Julia has [rules about what you can do inside of a finalizer](https://docs.julialang.org/en/v1/base/base/#Base.finalizer). In particular, I/O is a no go, so I assume erroring is bad.
2. We never really provide useful information when `destruct` fails:
  https://github.com/roualdes/bridgestan/blob/6e50958cc5f2079fc332807619e9b6637e259f85/src/bridgestan.cpp#L36-L42
  In fact, I am pretty sure we can make `~bs_model()` a `noexcept` function. I believe it will never throw. So I might clean this up to avoid that entirely.

WardBrian: Confirmed with @bob-carpenter that our destructor will not throw, so this is dead code. Going to clean up now

julia/src/model.jl
```diff
@@ -564,3 +584,18 @@ function log_density_hessian(
     hess = zeros(dims * dims)
     log_density_hessian!(sm, q, grad, hess; propto = propto, jacobian = jacobian)
 end
+
+"""
+    handle_error(sm::StanModel, err::Ref{Cstring}, method::String)
```
roualdes, line 589:
```suggestion
    handle_error(lib::Ptr{Nothing}, err::Ref{Cstring}, method::String)
```

roualdes (approved):
Looks good to me.  Thanks!

Issue #93, opened 2023-03-27
How should we handle print statements in Stan code?


WardBrian:
Also forking off from #90.

Right now `print` statements in Stan end up getting sent to `std::cerr` unconditionally. Two things about this

1. If we're going to hardcode anything, we should probably make this `std:cout`
2. Some languages like Python have ways to redirect ""stdout"". I put this in quotes because Python's `sys.stdout` and C/C++'s `stdout` are not necessarily the same, the former does not even need to point to a valid file descriptor but could be a class. Unifying Stan's `print` statements with this language-specific behavior is tricky, though there are known tricks like what @bob-carpenter links to here: https://github.com/roualdes/bridgestan/issues/90#issuecomment-1485389152

This is in many ways trickier than error messaging, since it is much harder to have it be zero-cost when you don't need it (the various redirection tricks you can use in Python all have a cost even if you never print anything. One potential solution is to have it be opt-in behavior, where you could write code like:
```python
import sys
import bridgestan
m = bridgestan.StanModel(...) # initialize model

sys.stdout = # some file we want to redirect to
m.log_density(...) # would print to C's stdout, NOT that file
with bridgestan.redirect_stdout():
    m.log_density(...) # in this scope we WOULD redirect
```

WardBrian:
Note: Julia's behavior is strictly better than Python's here, since their built-in [`redirect_stdout`](https://docs.julialang.org/en/v1/base/io-network/#Base.redirect_stdout) function does redirect C streams:
>Create a pipe to which all C and Julia level [stdout](https://docs.julialang.org/en/v1/base/io-network/#Base.stdout) output will be redirected

So without us doing anything, print statements in Stan work just like print statements in Julia natively.

Python's [library function of the same name](https://docs.python.org/3/library/contextlib.html?highlight=contextlib#contextlib.redirect_stdout) does not behave the same, so there would be different behavior between print statements inside Stan and those inside Python.

WardBrian:
Related: Any handling (including Julia's mentioned above) is predicated on Stan's print statements properly flushing the stream they're printing to. Currently they do not, see https://github.com/stan-dev/stanc3/pull/1301

If for some reason that PR does not go through, we could flush cout as part of BridgeStan, but I'd like to avoid that if at all possible.

aseyboldt:
It might also be an option to allow users of the C Api (ie the language bindings) to add a function that handles messages that stan want to print.

[...]

PR #94, opened 2023-03-28
Send Stan outputs to stdout rather than stderr

WardBrian:
Partially addresses #93.

Currently, all Stan print statements were redirected to stderr, this changes to stdout while we explore other options.

### DIFF
--- a/src/model_rng.cpp
+++ b/src/model_rng.cpp
@@ -62,7 +62,7 @@ bs_model_rng::bs_model_rng(const char* data_file, unsigned int seed,
   std::string data(data_file);
   if (data.empty()) {
     auto data_context = stan::io::empty_var_context();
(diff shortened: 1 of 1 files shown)


roualdes (approved):
So these changes are only for Stan print statements, ya? Which is why the error messages in bridgestan.cpp are left unchanged, right?

And printing to cout still depends on Stan flushing (re https://github.com/stan-dev/stanc3/pull/1301)?  Since that PR has already been approved, this PR is just moving a touch faster?  Which I'm ok with; just trying to make sure that a second set of eyes crosses t's and dots i's.

WardBrian:
Correct, this only affects Stan print statements. 

I just merged the stanc3 PR to flush the output, but even without that this is still better than the current state of things

Issue #95, opened 2023-03-28
Distinguish in docs BridgeStan Users from src C-API callers


roualdes:
It seems like the greater BridgeStan community could benefit from setting, and using consistently within our doc, specific nouns to distinguish between BridgeStan users and callers of the C level API.  

As I think of it, users refers to people who touch the interface level API.  Further, in my opinion, we don't have a relatively settled noun for those who touch exclusively the C level API, distinct from the interface API. 

Do you all agree such a distinction would help?  If so, what nouns do you suggest?

WardBrian:
I don't think there are many places in our documentation as it currently stands where the distinction is particularly strong right now, but I can imagine there being more over time. Right now the ["How it works"](https://roualdes.github.io/bridgestan/internals.html) page uses "developers" as a proxy for the second group, but that doesn't fit perfectly either.

This is why I was going back and forth with myself on whether or not #91 would require a major version bump. It's only really a breaking change if you're using the C API outside of the "official" wrappers we provide. We could go ahead and say that the C API is a moving target and things like extra arguments will be added without a major version bump if we can handle them in a backwards-compatible way in our interfaces, but that has its own downsides, especially if we imagine a world where interfaces exist outside our repo (if we decide Rust #88 should stay an external effort, for example)


I think the only person I have heard of who is using BridgeStan at the C level is @jtassarotti, who might have some thoughts on this.

bob-carpenter:
I think we should bump if we change the C API.  The client relationships in BridgeStan are:
  
| API | Client of | Description |
|:--|:--|:--|
| `src/bridgestan.cpp` |  | Core C++ API |
| `src/bridgestan.h` | `src/bridgestan.cpp` | Core C API |
| `src/bridgestanR.cpp` |  `src/bridgestan.cpp` | R-specific C++ API |
| `src/bridgestanR.h` | `src/bridgeStanR.cpp` | R-specific C API |
| `python/` | `src/bridgestan.h` | Python API |
| `julia/` | `src/bridgestan.h` | Julia API |
| `R/` | `/src/bridgestan.h` | R API |

WardBrian:
> Do we bump all the API's versions at once?

We have thus far. I really like this system, since it means "I have Bridgestan version x.y.z" installed has one, unambiguous meaning, but it makes things like major version bumps somewhat confusing if the breaking change doesn't affect the API in question.

#92 would likely require a breaking change in the interface-level code, so the next version would be a major version regardless of the C-API-only changes

[...]

PR #101, opened 2023-04-03
Expose version numbers in C API

WardBrian:
Mentioned by @aseyboldt in #95. This adds three declarations to `bridgestan.h`:

```C
extern int bs_major_version;
extern int bs_minor_version;
extern int bs_patch_version;
```

Which are populated by the version number macros we already have in version.hpp. Previously, these versions were only available by pulling apart the string returned by `bs_model_info()`.

This also adds a warning to the Python interface if the model version and python library version differ. This should be rare, but good to check. We can do something similar in Julia in 1.9+ when it is released: [`pkgversion`](https://github.com/JuliaLang/julia/pull/45607)

### DIFF
--- a/R/R/bridgestan.R
+++ b/R/R/bridgestan.R
@@ -41,6 +41,12 @@ StanModel <- R6::R6Class("StanModel",
         stop("Could not construct model RNG.")
       }
       private$model <- ptr_out
+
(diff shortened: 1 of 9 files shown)


roualdes (approved):
Looks good to me.  Thanks.

PR #102, opened 2023-04-05
WIP Rust interface for bridgestan

aseyboldt:
This is based on #88 and @WardBrian's [branch](https://github.com/WardBrian/bridgestan/tree/examples/rust-example).
Depends on #98 and will need to be rebased when that is merged.

It is still not decided if this should live in nutpie or in bridgestan, either would be fine for me.

I made a couple of changes compared to Brians branch:

- The shared library is stored as a `Borrow<StanLibrary>`, so that client code can decide what kind of reference should be used to store the Library (ie Rc, Arc, &lib or direct ownership). That avoids a whole range of headache when using this in multithreaded code.
- open_library checks that the version of the library has the same minor and major version as the rust library.
- `Model` is now Send and Sync
- STAN_THREADS=true is required
- Some more functions are wrapped
- If there is non-utf8 in an error message, it will now still return the message, but replace the invalid characters by placeholders.

Some open questions:

I'm not 100% sure about the handling of string/bytes data in the interface. It might be easier to return &CStr in a couple of places, and leave worrying about the encoding to users.

param_unconstrain: What lengths are allowed for theta? I'm assuming that include_tp, inclued_gq are both set to false right now...


WardBrian:
I've changed the base branch just to make viewing the diff easier, it would obviously still be merged into `main` ultimately

WardBrian:
TIL about `Borrow<T>`. Now I need to learn why one would prefer that over `AsRef<T>`.

For string encoding, I think you need to be in a really weird edge case to end up with a mangled error string. All of the "normal" Stan error messages should be plain ASCII, and I think any libraries we use in C++ would be the same. I'm personally happy to do the encoding for now and re-assess if we ever find a case it doesn't work in.

aseyboldt:
> TIL about Borrow<T>. Now I need to learn why one would prefer that over AsRef<T>.

The reason I wanted `Borrow` is this impl (which isn't available for `AsRef`):
https://doc.rust-lang.org/std/borrow/trait.Borrow.html#impl-Borrow%3CT%3E-for-T

This makes it possible to pass in the library directly (and not a reference at all), so that the model owns the library.

aseyboldt:
@WardBrian I guess this got closed because the branch was deleted?

WardBrian:
Oh, yeah, forgot that would happen!

this will need some rebasing anyway because of things like the constructor function names changing

aseyboldt:
I just did, I can just put it into a new PR or someone reopens this one (I can't)

WardBrian:
I also cannot reopen this. I guess there’s no way to change the base after the base has been deleted

Issue #103, opened 2023-04-06
Some more possible C API issues / suggestions


aseyboldt:
While working on rust langue support I found a few more (somewhat nitpicky) issues with the C API:

## File paths

`bs_construct` takes a file path as one input (as a `const *char`). It then passes that to `std::ifstream`. I spend quite some time yesterday trying to figure out how different operating systems handle encodings for filenames and paths, and it seems to be quite a mess. I'm really not sure how I should pass in the path on different operating systems so that it works with non ASCII paths.
Is that really worth the effort to figure that out? I think it would be easier to specify that the input should always be a utf8 encoded json c-string. (Or change it to `const *char` + strlen?).
Language bindings can then use their languages mechanism to handle file paths, so the C API doesn't have to worry about that at all.

It is also a bit unclear right now what I should do if there is no input data. I think I have to pass in an empty string, but I think a null pointer would be cleaner?

## Access to transformed data

I think there is currently no way to access quantities from the `transformed data` block? It would be nice if there was a way to extract those, so that they could for instance end up in the `data` group of an InferenceData object.

## Integer generated values

If I understand the type system of stan correctly, it has primitive types `float64` and `int32`.
In all functions of the C API that handle parameters, everything is a float64. I think that should be lossless, because every i32 value can be exactly represented as one float64 value.
I am not clear how complex values are encoded in the variable names though, and it would be great to have a way to figure out the actual types of the variables, so that I can convert them later in the traces.

WardBrian:
> File paths

We already optionally support c-string encoded JSON. The argument name is a bit historical/misleading, but the API doc specifies:
```
 * @param[in] data_file C-style string. This is either a
 * path to JSON-encoded data file (must end with ".json"), or
 * a JSON string literal.
```

It's not stated explicitly here, but Stan uses the [`rapidjson`](https://rapidjson.org/md_doc_encoding.html) parser which should be UTF-8 aware (we pass the `kParseValidateEncodingFlag` flag)

> Access to transformed data

These are stored as members of the generated model class object and are not accessible through the `model_base` interface (or, more generally, not accessible by any code which does not know the names ahead of time).

It's a reasonable thing to request but can't be done without changes to things upstream in stan-dev/stan

> Integer generated values

[...]

PR #104, opened 2023-04-06
Allow nullptr as data argument to constructor

WardBrian:
This was requested in #103.

Currently, you can specify a model has no data by passing "". This also allows you to pass `nullptr` for the same thing.

### DIFF
--- a/src/bridgestan.cpp
+++ b/src/bridgestan.cpp
@@ -16,8 +16,8 @@ bs_model_rng* bs_construct(const char* data, unsigned int seed,
   } catch (const std::exception& e) {
     if (error_msg) {
       std::stringstream error;
-      error << "construct(" << data << ", " << seed << ", " << chain_id
-            << ")"
+      error << "construct(" << (data == nullptr ? "NULL" : data) << ", " << seed
+            << ", " << chain_id << ")"
             << " failed with exception: " << e.what() << std::endl;
       *error_msg = strdup(error.str().c_str());
     }
@@ -25,8 +25,8 @@ bs_model_rng* bs_construct(const char* data, unsigned int seed,
     if (error_msg) {
       std::stringstream error;
 
-      error << "construct(" << data << ", " << seed << ", " << chain_id
-            << ")"
+      error << "construct(" << (data == nullptr ? "NULL" : data) << ", " << seed
+            << ", " << chain_id << ")"
             << " failed with unknown exception" << std::endl;
       *error_msg = strdup(error.str().c_str());
     }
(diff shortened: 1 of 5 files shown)


src/bridgestan.cpp
```diff
@@ -9,14 +9,14 @@ int bs_patch_version = BRIDGESTAN_PATCH;
...
+bs_model_rng* bs_construct(const char* data, unsigned int seed,
                            unsigned int chain_id, char** error_msg) {
   try {
-    return new bs_model_rng(data_file, seed, chain_id);
+    return new bs_model_rng(data, seed, chain_id);
   } catch (const std::exception& e) {
     if (error_msg) {
       std::stringstream error;
-      error << "construct(" << data_file << ", " << seed << ", " << chain_id
+      error << "construct(" << data << ", " << seed << ", " << chain_id
```
aseyboldt, line 19: I guess this is invalid if `data` is `NULL`?

WardBrian: It seems that you can use `operator<<` with a null pointer but it terminates the string. E.g, you get an error which looks like:
`Error: construct(` \<EOL\>
which isn't great. 

It's easy to special-case, though.

aseyboldt: Sorry for all the pedantic comments recently.
Seems I get that way when I write too much rust. :-)

WardBrian: Absolutely no need to apologize, I appreciate them. Every pedantic comment now is a future bug report -> pr -> patch release cycle we get to skip.

(coincidentally, this is why I like languages like Rust or OCaml with picky compilers. The downside is sometimes it ends up with me being far too trusting of my C++ compiler...)

bob-carpenter:
@aseyboldt:  You picked the right team for pedantry.  We really do appreciate all the detailed comments.

src/bridgestan.h
```diff
@@ -18,9 +18,9 @@ extern int bs_patch_version;
  * generator (PRNG) wrapper.  Data must be encoded in JSON as
  * indicated in the *CmdStan Reference Manual*.
  *
- * @param[in] data_file C-style string. This is either a
- * path to JSON-encoded data file (must end with ".json"), or
- * a JSON string literal.
+ * @param[in] data C-style string. This is either a
+ * path to JSON-encoded data file (must end with ".json"),
+ * a JSON string literal, or nullptr, which is interpreted as no data.
```
roualdes, line 23:
```suggestion
 * a JSON string literal, an empty string, or nullptr, which is interpreted as no data.
```

roualdes (approved):
Just one doc suggestions.  Otherwise, looks good to me.

src/bridgestan.h
WardBrian: Agreed, it is good to call out this case specifically.

PR #106, opened 2023-04-18
Utilize doc version switcher

WardBrian:
This is something I've been thinking about since we merged #91 and #98 but haven't yet had a release: our online docs have updated to the latest API, but people who download/use it right now won't have that necessarily. 

Obviously one fix is to always be prompt on releasing new code, but I think a better solution involves a more complete documentation site. Our theme supports [version switchers](https://pydata-sphinx-theme.readthedocs.io/en/stable/user_guide/version-dropdown.html), which we can use to keep multiple versions of the docs around.

This PR does a few things:

1. The docs built on every push to `main` will now live at `roualdes.github.io/bridgestan/latest/`
2. Each release will create a new sub-page, e.g. for version `2.0.1` there will be a copy at `roualdes.github.io/bridgestan/v2.0.1/`
3. Our docs will have a nice version switcher to go between these different versions.

This will all be automated going forward, but the first time through I will need to make some manual changes to the `gh-pages` branch.

1. We should add a redirect from `roualdes.github.io/bridgestan/` to `roualdes.github.io/bridgestan/latest/` (easy)
2. It would be nice to go "back in time" and manually build a couple older versions of the docs with this enabled.


roualdes (approved):
Great idea.  The changes look good to me, in my novice github workflow opinion.

WardBrian:
Updates should be live. https://roualdes.github.io/bridgestan will redirect to https://roualdes.github.io/bridgestan/latest, and the new drop down is visible in the top right

PR #108, opened 2023-04-18
Add the ability to provide a callback function for printing

### LINKED ISSUE #93: How should we handle print statements in Stan code?

WardBrian:
Closes #93 with a solution based on the discussion in there.

This PR adds a function `bs_set_print_callback` which accepts a pointer to a function with the C signature 
`void callback(const char* text, size_t length)`. This is then used to construct a custom `ostream` used in place of `std::cout` in output generated inside the Stan program.

For languages like Python which don't play very nice with the C-level `stdout`, this allows output capturing by passing in a function which can in turn call Python's `print()` function. This isn't necessary for Julia, which could already capture output from C calls.

**Efficiency**:

Note that the custom ostream we create does use a mutex to ensure multiple threads can each use the callback safely. This, as well as the cost of a callback which is running Python code, could result in noticeable slowdowns when printing from inside Python. **However**, this will only matter for programs with print statements, which would primarily mean they are being debugged or that speed is not the primary concern.

Users who are really concerned about this can turn it off by passing `capture_stan_prints=False` to the constructor in Python.


WardBrian:
The C++ changes (primarily in `src/callback_stream.hpp`) are based off some digging in the C++ doc and [an old blogpost](http://videocortex.io/2017/custom-stream-buffers/)

It seems to work well, but I must admit I'm not terribly confident in it. Some extra scrutiny on that code would be great!

roualdes (changes requested):
Awesome work, Brian.  Thanks much.

I gave this an honest look over, but this is at the limits of my knowledge of C++.

A few questions

1. We're not using the C++ proposed here for Julia.  Julia "just works" here when single threaded.  Does it work when multithreaded?  More specifically, are the Julia stdio redirects thread safe?  Is it worth a test to check?
2. There are rumors about Python's `print(..., end = "\n")` flushing automatically when the keyword argument `end = "\n"`, though I couldn't confirm this.  This shouldn't matter by Stan 2.32, if I understand things correctly.  Are we waiting until Stan 2.23 to merge this, or should we doc that users should use `print(..., flush = True)`?
3. Should we add doc on how to use this?
4. Should we write a test that tests the mutex specifically?

WardBrian:
> Aren't we sort of just testing the C++ standard library at that point, though?

I said that and then immediately got unexpected behavior when I tried it. Looking into it now

WardBrian:
Okay, re: testing the mutex

- I first tried doing this in Python, like so:
  ```python
    import ctypes, threading
    x = 0
    @ctypes.CFUNCTYPE(None, ctypes.c_char_p, ctypes.c_int)
    def callback(s, n):
        nonlocal x
        x += 1

    m2._set_print_callback(callback, None)
    def f():
        rng = m2.new_rng(1234)
        for _ in range(5):
            m2.param_constrain(np.array([theta]), include_gq=True, rng=rng)
    threads = [threading.Thread(target=f) for _ in range(10)]
    for t in threads:
        t.start()

    assert x == 2 * 5 * 10
  ```
  This _did not work_, x was all over the place (usually smaller). However, the ctypes doc has this note:

  > ctypes creates a new dummy Python thread on every invocation. This behavior is correct for most purposes, but it means that values stored with [threading.local](https://docs.python.org/3/library/threading.html#threading.local) will not survive across different callbacks, even when those calls are made from the same C thread.
- So, worried that it was actually just a poor test in Python, I wrote some C code:
  ```C
  #include "bridgestan.h"
  #include <stdio.h>

  #include <pthread.h>

  void bs_print(bs_model* m) {
    double theta[1] = {0.5};
    double theta_out;
    bs_rng* rng = bs_rng_construct(123, NULL);
    for (int i = 0; i < 25; i++)
      bs_param_constrain(m, 0, 1, theta, &theta_out, rng, NULL);
  }

  volatile int x = 0;
  void pcb(const char* msg, size_t len) { x += 1; }

  #define NUM_THREADS 10
  void print_test(bs_model* model) {
    bs_set_print_callback(pcb, NULL);

    pthread_t id[NUM_THREADS];
    for (int i = 0; i < NUM_THREADS; i++) {
      pthread_create(&id[i], NULL, (void*)bs_print, model);
    }

    for (int i = 0; i < NUM_THREADS; i++) {
      pthread_join(id[i], NULL);
    }
  }
  ```
  This, by contrast, is rock solid. As a bonus, removing the `lock_guard` from the `callback_ostreambuf` class _breaks_ the above code, so we know it is doing something. 

In conclusion: I believe we are doing the right thing in C++. It's not clear to me if Python's callbacks are very robust to threading, but almost none of Python _is_, so that's not too shocking. 

I'm not sure how far we should (or even can?) push on the Python side

roualdes:
Thanks, Brian, for giving potential thread safety issues such a thorough investigation.  Glad to hear that Python is likely the culprit for threading issues and not the C++ code.

> What additional doc would you suggest?

This was in part based on a mistake in my thinking, because indeed

> this should just silently do the right thing

Further, I suppose we can assume that users of BridgeStan in general know that printing is relatively slow, so that we don't need to tell them this in our doc.

Just tried this out in Jupyter for both Python and Julia.  This works as expected in Python, given what we learned from above.  

From what I can tell Stan print statements are buffered when going through Julia.  Maybe you knew this already, hence `Base.libc.flush_cstdio` in the Julia tests.  I'm just double checking things for my own sanity. Will Stan 2.23 fix this or is this a Julia thing?

I'll go ahead approve now.  Thanks for double checking some things.

roualdes (approved):

WardBrian:
> Will Stan 2.32 fix this or is this a Julia thing?

Yes, I updated the generated code by the compiler in https://github.com/stan-dev/stanc3/pull/1301 to terminate prints with `std::endl` which flushes

WardBrian:
@roualdes embarrassing error: if you compare my python code above to my C code above, you'll notice the Python code is missing a loop which calls `thread.join()` on each thread. 

This, it seems, was entirely the source of the issue with the Python code. So, adding a test is reasonable after all.

roualdes (approved):
Excellent.  Thanks for this great PR.

PR #109, opened 2023-04-20
Update to Stan 2.32

WardBrian:
This updates to the newly-released Stan 2.32.0 (see https://github.com/stan-dev/cmdstan/issues/1154)

The only change which are particularly relevant to BridgeStan are the flushing of print statements (see #108) and the changes to `get_dims` and `get_param_names` allowing us to delete a decent chunk of code we don't need


roualdes (approved):
Buffered Stan print statements via Julia was short lived. Very nice.  Looks good to me.

PR #112, opened 2023-04-25
Read in JSON inside higher-level languages

WardBrian:
As discussed in #103, we may benefit from some additional engineering effort in terms of path handling, etc if we load the JSON files for data inside the higher level language and pass the resulting string.


roualdes (approved):
Ya, this seems fine to me.

Issue #117, opened 2023-05-05
Python ctypes.CDDL compiled binary loading error


Dennis8:
I'm trying to get bridgestan to work on Windows 10 Version 10.0.19045 Build 19045.
I've downloaded both the Pypi version (through pip install) and the git clone version.
I can compile the .stan code to a .so binary and .hpp file with `make test_models/bernoulli/bernoulli.stan`.
However the compiled binary will not load into Python with the class StanModel:

```python
import bridgestan as bs
import numpy as np

stan = "../test_models/bernoulli/bernoulli.stan"
data = "../test_models/bernoulli/bernoulli.data.json"

model = bs.StanModel.from_stan_file(stan, data)

print(f"This model's name is {model.name()}.")
print(f"It has {model.param_num()} parameters.")

x = np.random.random(model.param_unc_num())
q = np.log(x / (1 - x))  # unconstrained scale
lp, grad = model.log_density_gradient(q, jacobian=False)
print(f"log_density and gradient of Bernoulli model: {(lp, grad)}")
```

```text
Traceback (most recent call last):
  File "...\bridgestan_test\example.py", line 26, in <module>
    model = bs.StanModel.from_stan_file(stan, data)
  File "...\venv\lib\site-packages\bridgestan\model.py", line 244, in from_stan_file
    return cls(
  File "...\venv\lib\site-packages\bridgestan\model.py", line 71, in __init__
    self.stanlib = ctypes.CDLL(self.lib_path)
  File "...\Python310\lib\ctypes\__init__.py", line 374, in __init__
    self._handle = _dlopen(self._name, mode)
FileNotFoundError: Could not find module '...\.bridgestan\bridgestan-2.0.0\test_models\bernoulli\bernoulli_model.so' (or one of its dependencies). Try using the full path with constructor syntax.
```

I have tried to trace where it goes wrong and the following also doesn't seem to work:
```python
from pathlib import Path
BS_FILEPATH = r"C:\Users\MyUsername\.bridgestan\bridgestan-2.0.0\python"
compiled = Path(BS_FILEPATH).parent / "test_models/bernoulli/bernoulli_model.so"
print(compiled.exists()) # Returns True

import ctypes
stanlib = ctypes.CDLL(str(compiled))
```

FileNotFoundError: Could not find module '...\.bridgestan\bridgestan-2.0.0\test_models\bernoulli\bernoulli_model.so' (or one of its dependencies). Try using the full path with constructor syntax.

[...]

Issue #100, opened 2023-04-03
Ability to build models from inside Rust


WardBrian:
We have this ability in Julia, Python, and R, but not Rust.

PR #118, opened 2023-05-05
Windows: add TBB (and MinGW DLLs) to load path

WardBrian:
This is in response to #117. It isn't a full "fix" (arbitrary user configs could still run into similar problems and need to set their own `%PATH%`s correctly), but it attempts to be a bit friendlier.

This is similar to code in [cmdstanpy](https://github.com/stan-dev/cmdstanpy/blob/cd650840d9b8dec02b98d563bcdf88ccefb10fff/cmdstanpy/model.py#L216-L223). The main difference is (in Python) it also uses [`os.add_dll_directory`](https://docs.python.org/3/library/os.html#os.add_dll_directory). This appears to have fixed the issue which was preventing us from running our Python tests on Windows in Github Actions, so I've re-enabled those.

I'll do a follow on PR which tries to split the Windows CI in to separate jobs per-interface. 

Note: This PR does not change the R interface, since R doesn't (yet) have any handling of where BridgeStan is installed (see #100)

### DIFF
--- a/julia/src/BridgeStan.jl
+++ b/julia/src/BridgeStan.jl
@@ -43,4 +43,17 @@ StanModel(;
     seed = 204,
 ) = StanModel(compile_model(stan_file; stanc_args, make_args), data, seed)
 
+
+function __init__()
+    # On Windows, we may need to add TBB to %PATH%
+    if Sys.iswindows()
+        try
+            run(pipeline(`where.exe tbb.dll`, stdout=devnull, stderr=devnull))
+        catch
+            # add TBB to %PATH%
+            ENV["PATH"] = joinpath(get_bridgestan_path(), "stan", "lib", "stan_math", "lib", "tbb") * ";" * ENV["PATH"]
+        end
+    end
+end
+
 end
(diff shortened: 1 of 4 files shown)


WardBrian:
It turns out that whether the os.add_dll_directory call is necessary is OS (windows only), version (behaves differently in 3.8 and 3.10), AND distribution (anaconda python doesn’t need it, python from python.org does) specific. 

I have confirmed that it never hurts in situations where it isn’t required, but that explains why I could never get the tests to run in CI but could locally.

julia/src/BridgeStan.jl
roualdes, line 1: The logic here seems slightly different, as compared to the logic in Python.  So this comment is just meant to double check on that difference.

Under Julia, if `where.exe` can find `tbb.dll`, then the Julia `ENV["PATH"]` will be able to successfully find `tbb.dll`, is this correct?  And it's only if `where.exe` can't find `tbb.dll` that we need to add `tbb.dll` to the path?

roualdes (approved):
Just my one question, which you likely already thought through.  I just want to make sure I'm on the same page.  Thanks.

julia/src/BridgeStan.jl
WardBrian: Yes, Python specifically does not look in PATH for dependent DLLs, which as far as I can tell all other languages do.

PR #119, opened 2023-05-06
Split Windows CI

WardBrian:
This is based on #118 and should be merged second. Dramatically cleans up our CI action and makes Windows tests more similar to the other platforms.

### DIFF
--- a/.github/workflows/main.yaml
+++ b/.github/workflows/main.yaml
@@ -31,18 +31,18 @@ jobs:
           path: ./stan/
           key: ${{ runner.os }}-stan-${{ hashFiles('stan/src/stan/version.hpp') }}-v${{ env.CACHE_VERSION }}
 
-      # - name: Build C example (Unix)
-      #   if: matrix.os != 'windows-latest'
-      #   run: |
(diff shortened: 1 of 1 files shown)


roualdes:
This PR after #118, check.  After a cursory look, just one question.  What do you think about adding Python 3.11 and Julia 1.8 to our test matrix?

roualdes:
> unless you think we could test on 3.9 and 3.11, skipping 3.10

I like the idea of testing an earlier version, namely 3.9, and the latest release, 3.11, but I'm not sure how much this preference corresponds with any tangible aspect of reality.  For instance, just because we test at some made up ends of this range of Python versions doesn't really guarantee that everything in between works, e.g. re #118 os.add_dll_directory is version dependent.

Further, I have no sense of how frequently people stay up to date on their Python versions.

So I'm not sure what ditching 3.10 in favor of 3.11 gets us.  I could go either way on this one.

roualdes (approved):
This PR looks good to me.  The cleaned up workflow is nice.  Thanks.

PR #122, opened 2023-05-11
Fix R example and run examples in CI

### LINKED ISSUE #121: Error in R/example.R
When I run the R example I got this error 

`Error in initialize(...) : unused argument (0) `

From the function help I think the `0` is an extra argument here  

https://github.com/roualdes/bridgestan/blob/32bd072236a0b59e52dfeb69c0a0b7342d6d7b42/R/example.R#L7

WardBrian:
This closes #121 by removing the extra argument (left over from pre- #98) and also adds a line to run each example program as part of our CI to make sure they stay up-to-date.

### DIFF
--- a/.github/workflows/main.yaml
+++ b/.github/workflows/main.yaml
@@ -105,7 +105,9 @@ jobs:
 
       - name: Run tests
         run: |
-          pytest -v python/
+          cd python/
+          python example.py
+          pytest -v
         env:
           BRIDGESTAN: ${{ github.workspace }}
 
@@ -155,7 +157,9 @@ jobs:
 
       - name: Run tests
         run: |
-          julia --project=./julia -t 2 -e "using Pkg; Pkg.test()"
+          cd julia/
+          julia --project=. example.jl
+          julia --project=. -t 2 -e "using Pkg; Pkg.test()"
         env:
           BRIDGESTAN: ${{ github.workspace }}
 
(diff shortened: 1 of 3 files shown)


bob-carpenter:
The PR looks OK, but there's still a failing test on the Julia side.

WardBrian:
I can’t figure out the Julia test failure. My current guess is it’s an issue with Julia 1.9.0 which was released yesterday

roualdes:
I've got to move on to other things for now, but here's a quick update on my progress so far.  Julia 1.8 and 1.9 output different error messages when tested against the same faulty Stan programs.

* Load the Stan program `test_models/throw_data/throw_data.stan`:

Julia 1.8: ERROR: construct(, 204) failed with exception: Exception: find this text: datafails

Julia 1.9: ERROR: construct(, 204) failed with unknown exception

* Run `BridgeStan.log_density(model, rand(1))` with Stan program `test_models/throw_lp/throw_lp.stan`:

Julia 1.8: ERROR: log_density() failed with exception: Exception: find this text: lpfails

Julia 1.9: ERROR: log_density() failed with unknown exception

What seems odd to me is the differences between these error messages comes from the C++ file `bridgestan.cpp`.  The C++ try-catch is catching different exceptions based on the Julia version.  I don't immediately understand this.

The quickest/hackiest fix would be to change line 9 in the file `julia/test/model_tests.jl` from `if VERSION >= v"1.8"` to `if VERSION == v"1.8"`.  Is this a reasonable temporary fix, while we create a new issue focused on Julia 1.9 and move away from this Rlang PR?

WardBrian:
I've also spent some time running it down and just filed an issue with Julia: https://github.com/JuliaLang/julia/issues/49772. 
I believe this is a legitimate bug in Julia that has nothing to do with BridgeStan specifics. I'm happy to either bin Julia 1.8 in our test suite until it is resolved or merge with the one test failing.

Issue #84, opened 2023-03-20
remaining autodiff functors


bob-carpenter:
There are a bunch of autodiff functors that are implemented in Stan but not exposed yet in BridgeStan.  The two most basic are already done.   Most of them other than directional derivatives require forward-mode autodiff.   Please feel free to add more requests to the list.

* [x] gradient
* [x] Hessian
* [ ] 3rd order derivatives (called `grad_Hessian` in `stan/math/mix`; requires forward mode)
* [ ] directional derivatives (called `gradient_dot_vector` in `stan/math/mix`; most efficient in forward, can do backward)
* [ ] Hessian-vector product (called `hessian_times_vector` in `stan/math/mix`; requires forward mode)
* [ ] gradient of the trace of Hessian times a matrix, which is required for softabs Riemannian (`grad_tr_mat_times_hessian` in `stan/math/mix`; requires forward mode)

There's no inverse Hessian vector product in Stan.  I'm not sure the best way to implement that---I think there are a lot of approaches because the direct way is so expensive (`vector / Hessian` in Stan).

bob-carpenter:
There are two versions of Hessian.  

1. `stan/math/mix/functor/hessian.hpp` requires `fvar<var>` and does N evals for an N-dimensional density.

2. `stan/math/fwd/functor/hessian.hpp` requires `fvar<fvar<double>>` and does N^2 evals of an N-dimensional density.

We want to use approach (1).   They implement the same interface, so it's just a matter of changing the include.

WardBrian:
`grad_hessian` as currently implemented doesn't seem to actually call either version of `hessian` directly, so it will probably be a bit more invasive of a change than just an include. Still an upstream change, just not the one I was originally imagining (and probably an easier one than `fvar<fvar<var>>` for `log_prob`)

It seems to me like third-order autodiff should require that extra level of nesting? But I may be imagining this incorrectly.

WardBrian:
We must be talking past each other:

- BridgeStan currently can call either the finite diff hessian or the "version 1" (`fvar<var>`) function you mention above. No issues here, I only mentioned it to say that supporting the `fvar<var>` version required upstream changes

- My original comment was that `grad_hessian` cannot be exposed without *further* changes to upstream Stan functionality due to the requirement for `fvar<fvar<var>>`, which sounds like it is correct

bob-carpenter:
> we can only call overloads which exist in the model_base_crtp

[...]

PR #126, opened 2023-05-30
Feature: Hessian-vector product function

### LINKED ISSUE #125: Hessian-vector product from stan-dev/math
Expose the version of the Hessian-vector product functional from [`stan-dev/math`](https://github.com/stan-dev/math).

Here's the functional we need to expose.  We want the first one that uses reverse-mode nested in forward mode):

https://github.com/stan-dev/math/blob/develop/stan/math/mix/functor/hessian_times_vector.hpp

We should use the first version in the file (reverse nested in forward), because it's linear.  Doing it with finite diffs and multiply will be quadratic (as will be using the second version in the source file).

The motivation is the optimizer (conjugate gradient trust region method) used by the following paper, which uses [`trust-ncg`](https://docs.scipy.org/doc/scipy/reference/optimize.minimize-trustncg.html) in `scipi.optimize.minimize`:

https://arxiv.org/pdf/2304.05527.pdf

@rgiordan said that they used PyMC because it provided Hessian-vector products.

WardBrian:
Closes #125 and completes part of #84.

This adds a function with the following signature to the C API:
```C
int bs_log_density_hessian_vector_product(const bs_model* m, bool propto,
                                          bool jacobian,
                                          const double* theta_unc,
                                          const double* vector, double* val,
                                          double* Hvp, char** error_msg);
```

When `BRIDGESTAN_AD_HESSIAN` is defined, this uses `stan::math::hessian_vector_product`. Otherwise, I currently have it implemented to compute the hessian and then just multiply it by the supplied vector. 

In Stan 2.33 we can use https://github.com/stan-dev/math/pull/2914


bob-carpenter:
I would prefer to default to a system that works rather than one that throws errors.  A first implementation can build the entire Hessian with finite diffs of gradients, then multiply.   A less memory intensive version would generate one row of the Hessian at a time with finite diffs of gradients, then multiply to get a single entry in the result.  This is, of course, parallelizable, as would be the Hessian algorithm in AD for that matter.

aseyboldt:
In the case of using finite diffs, why not approximate $H_pv$ by $\frac{\nabla f(p + hv) - \nabla f(p)}{h}$ (after normalizing $v$)? Is there a problem with that I should be aware of?

bob-carpenter:
We're doing the central version (second thing you wrote under "related finite diff").  It's implemented as a functional in Stan's C++ here and we just plugged that into BridgeStan:

https://github.com/stan-dev/math/blob/develop/stan/math/rev/functor/finite_diff_hessian_auto.hpp

aseyboldt:
Sorry if I'm missing something here, but doesn't that compute the complete hessian (in 2n grad calls). I was referring to the finite diff version of a hessian vector product (with 2 grad calls).

bob-carpenter:
Nope---I was the one missing the point.  Yes, what I pointed to computes the complete Hessian.  

I hadn't though about updating the finite diff algorithm to do Hessian-vector products innately.  The place to add that would probably be the Stan math lib.  It'd be faster (and more accurate?) than what we're doing now for tests, which I'm pretty sure just computes the Hessian and multiplies.

WardBrian:
Exposed this function in each interface. For the moment I've left it as the naive "calculate the hessian via finite differences and multiply" approach when autodiff hessians are unavailable.

bob-carpenter:
> I've left it as the naive "calculate the hessian via finite differences and multiply" approach

I think that's OK for a first implementation.  

@aseyboldt :  do you know how to write the algorithm to do this more efficiently with finite differences over gradients that doesn't involve computing the whole Hessian?

aseyboldt:
I guess pretty much this one:

```
from scipy import optimize, linalg

# From scipy.optimize.approx_fprime
_epsilon = np.sqrt(np.finfo(float).eps)

def hessian_vector_prod_approx(grad_func, x, v, *, h=_epsilon):
    v_norm = linalg.norm(v)
    v = v / v_norm
    grad_forward = grad_func(x + h*v)
    grad_backward = grad_func(x - h*v)
    return v_norm * (grad_forward - grad_backward) / (2 * h)
```

The problem is that I really don't know too much about the numerical issues that I'm pretty sure are buried here, and just googling a bit for literature about this I couldn't find too much that looked useful...

bob-carpenter:
Thanks for the algorithm and reference.  @justindomke was visiting, but we can still check with him.  @bgoodri may also know---I think the central finite diffs rather than one-sided versions was his idea.

The place to code the faster version is in the `stan-dev/math` repo.  It should also be easier to test inside of the math repo.   Then we can just include it in BridgeStan.  We can also fix the link.

justindomke:
Hi everyone! I have a few things to add that might be helpful.

One is that you can think of a Hessian-vector product as two composed gradient evaluations. (This doesn't change anything, but I find it a useful way to look at things.) Say you want to compute $\nabla^2 f(x) v$, where $x$ and $v$ are vectors and $f: R^n \rightarrow R$. Define $g(x) = \nabla f(x)$ to be the gradient and define $h(x) = g(x) \cdot v$. Then $\nabla^2 f(x) v = \nabla h(x)$.

I don't know if Stan can do 2-nd order derivatives like this, but here's a little demo using JAX:

```python
import jax
from jax import numpy as jnp

def f(x):
    "random nonlinear function"
    return (x[0] * jnp.sum(jnp.sin(x)))**x[1]

def naive_hvp(x,v):
    return jax.hessian(f)(x) @ v

def recursive_hvp(x,v):
    g = jax.grad(f)
    h = lambda x: g(x) @ v
    return jax.grad(h)(x)

x = jnp.array([1.1,2.2,3.3])
v = jnp.array([4.4,5.5,6.6])
```

produces the output

```
naive_hvp(x,v)=Array([ -6.506892, -44.439754, -32.04876 ], dtype=float32)
recursive_hvp(x,v)=Array([ -6.5068817, -44.43976  , -32.048767 ], dtype=float32)
```

If you can do it this way, you probably should. `recursive_hvp` has the same complexity as `f` (up to some constant independent of the length of `x`) and has no finite-different floating point errors.

Second, there's essentially an infinite number of finite difference schemes, with different tradeoffs between computational cost and accuracy. You can go beyond 1-sided or 2-sided differences to "4-sided" or higher. [Here's short post from me giving the basic idea](https://justindomke.wordpress.com/2017/04/22/you-deserve-better-than-two-sided-finite-differences/). This is for derivatives rather than Hessian-vector products, but the same idea applies. You could modify  @aseyboldt 's code above to use any of these.

In practice many people seem to find two-sided differences to be a reasonable compromise. They at least work "most of the time" whereas with one-sided differences will fall over for even slightly challenging problems.

Finally, if you're doing finite differences, don't neglect the problem of choosing the perturbation size. As far as I know, this is hard and there is no general scheme that's always guaranteed to work well. But probably there are heuristics that are better than always using numerical epsilon. I've had good luck using the heuristic $r = \sqrt{\epsilon}(1 + \vert x \vert_\infty) \vert \nabla f(x) \vert_\infty$ (where $r$ is the perturbation size and $\epsilon$ is numerical epsilon) which I first found in [this paper](https://doi.org/10.1016/j.cam.2008.12.024).

bob-carpenter:
Thanks for the explanation and refs, @justindomke.  Stan has a reasonable Hessian-vector product implemented for everything but our implicit functions.  For models using implicit functions, we want to provide finite differences.  I opened an issue in the math lib, which is where the implementation should go:

https://github.com/stan-dev/math/issues/2913

When we were discussing finite diffs, someone made exactly the same point about 2 diffs being a good compromise.

WardBrian:
I took a pass at the above in https://github.com/stan-dev/math/pull/2914, more or less a direct translation of what @aseyboldt had above. 

For BridgeStan, we have a couple choices in the meantime:
1. Hold off on the next release of Stan, and use the stan::math implementation assuming one gets merged
2. Include a copy of it ourselves for now
3. Use the naive approach of calculating the Hessian and then multiplying when second order autodiff is not available.

bob-carpenter:
I'm OK with any of those choices---they all have pros and cons.  I'm leaning slightly toward (2) or (3) just to get the API in place even if we change implementations later.

WardBrian:
>  I've had good luck using the heuristic $r = \sqrt{\epsilon}(1 + \vert x \vert_\infty) \vert \nabla f(x) \vert_\infty$ (where $r$ is the perturbation size and $\epsilon$ is numerical epsilon) which I first found in [this paper](https://doi.org/10.1016/j.cam.2008.12.024).

@justindomke It seems like that paper suggests a different heuristic (namely $r = \frac{\sqrt{\epsilon} ( 1 + \Vert x\Vert)}{\Vert v \Vert} $) in eq 5.2, am I missing the other one somewhere?

roualdes:
I too am OK with any of those choice, lean towards 2. or 3., and I vote 3. just because that's already written up in this PR.

R/R/bridgestan.R
```diff
@@ -287,6 +284,31 @@ StanModel <- R6::R6Class("StanModel",
...
+    #' Return the log density and the product of the Hessian
+    #' with the specified vector.
+    #' @param theta_unc The vector of unconstrained parameters.
+    #' @param v The vector to multiply the Hessian by.
+    #' @param propto If `TRUE`, drop terms which do not depend on the parameters.
+    #' @param jacobian If `TRUE`, include change of variables terms for constrained parameters.
+    #' @return List containing entries `val` (the log density) and `Hvp` (the hessian-vector product).
+    log_density_hessian_vector_product = function(theta_unc, v, propto = TRUE, jacobian = TRUE){
+      dims <- self$param_unc_num()
+      vars <- .C("bs_log_density_hessian_vector_product_R",
```
roualdes, line 298: Does this need an `NAOK = TRUE`?

justindomke:
> @justindomke It seems like that paper suggests a different heuristic (namely r=ϵ(1+‖x‖)‖v‖) in eq 5.2, am I missing the other one somewhere?

No, I'm sure you're right! I missed a division symbol above (fixed now), and also mine was for gradients. Yours looks good.

WardBrian:
With the release of Stan 2.33 this has been updated to use the finite diff Hessian-vector-product from stan::math if autodiff hessians are unavailable. I believe it's ready for review @roualdes @bob-carpenter

julia/test/model_tests.jl
```diff
@@ -550,24 +550,24 @@ end
...
     R = 1000
-    out_size = BridgeStan.param_num(model;include_tp=false, include_gq=true)
+    out_size = BridgeStan.param_num(model; include_tp = false, include_gq = true)
 
     # to test the thread safety of our RNGs, we do two runs
     # the first we do in parallel
     gq1 = zeros(Float64, out_size, R)
-    @Threads.threads for it = 1:nt
-        rng = StanRNG(model,seeds[it]) # RNG is created per-thread
+    Threads.@threads for it = 1:nt
```
roualdes, line 558: Good catch.  I definitely would have guessed `@Threads.threads` would have errored.

julia/test/model_tests.jl
```diff
@@ -651,6 +651,9 @@ end
     using LinearAlgebra
     @test isapprox(-Matrix(1.0I, D, D), hess)
 
+    v = rand(D)
+    lp, Hvp = BridgeStan.log_density_hessian_vector_product(model, y, v)
+    @test isapprox(-v, Hvp)
```
roualdes, line 656: Do we have `BRIDGESTAN_AD_HESSIAN` defined for one interface language and not defined for another language, such that both C paths are tested?

src/bridgestan.cpp
```diff
@@ -205,6 +205,34 @@ int bs_log_density_hessian(const bs_model* m, bool propto, bool jacobian,
...
+      std::stringstream error;
+      error << "log_density_hessian_vector_product() failed with exception: "
+            << e.what() << std::endl;
+      *error_msg = strdup(error.str().c_str());
+    }
+  } catch (...) {
+    if (error_msg) {
+      std::stringstream error;
+      error << "log_density_hessian_vector_product() failed with unknown "
+               "exception"
```
roualdes, line 228: Should there be another chevron here, `<<`?

src/bridgestan.h
```diff
@@ -262,7 +262,9 @@ int bs_log_density_gradient(const bs_model* m, bool propto, bool jacobian,
  * pointer `hessian` must have enough space to hold the Hessian.
  *
  * The gradients are computed using automatic differentiation.  the
- * Hessians are
+ * Hessians are computed using nested automatic differentiation if
+ * BRIDGESTAN_AD_HESSIAN is defined, otherwise they are computed
+ * using central finite differences.
```
roualdes, line 267: Should we add "...using central finite differences of gradient vectors at theta_unc."?

Is a the missing capitalization at the beginning of this sentence?

src/bridgestan.h
```diff
@@ -280,6 +282,38 @@ int bs_log_density_hessian(const bs_model* m, bool propto, bool jacobian,
                            const double* theta_unc, double* val, double* grad,
                            double* hessian, char** error_msg);
 
+/**
+ * Calculate the log density and the product of the Hessian with the specified
+ * vector for the specified unconstrain parameters and write it into the
```
roualdes, line 287: unconstrain*ed*

src/bridgestan.h
```diff
@@ -280,6 +282,38 @@ int bs_log_density_hessian(const bs_model* m, bool propto, bool jacobian,
...
+ * specified value pointer and Hessian-vector product pointer, dropping
+ * constants it `propto` is `true` and including the Jacobian adjustment if
+ * `jacobian` is `true`. Returns a return code of 0 for success
+ * and -1 if there is an exception executing the Stan program. The
+ * pointer `hvp` must have enough space to hold the product.
+ *
+ * Hessian-vector-products are computed using nested automatic
+ * differentiation if BRIDGESTAN_AD_HESSIAN is defined, otherwise
+ * they are computed using central finite differences.
```
roualdes, line 296: same question about central finite differences of gradient vectors at theta_unc

src/model_rng.cpp
```diff
@@ -306,3 +306,28 @@ void bs_model::log_density_hessian(bool propto, bool jacobian,
...
+  Eigen::VectorXd hvp_vec(N);
+
+#ifdef BRIDGESTAN_AD_HESSIAN
+  stan::math::hessian_times_vector(logp, params_unc, v, *val, hvp_vec);
+#else
+  stan::math::internal::finite_diff_hessian_times_vector_auto(logp, params_unc,
+                                                              v, *val, hvp_vec);
+#endif
+
+  Eigen::VectorXd::Map(hvp, N) = hvp_vec;
```
roualdes, line 332: <off-topic>
Is this because Stan Math doesn't have a signature for `stan::math::internal::finite_diff_hessian_times_vector_auto()`, nor `stan::math::hessian_times_vector()` which accepts `hvp` as `double*`?

We added an extra signature for gradient, https://github.com/stan-dev/math/pull/2743.  Maybe we should for `stan::math::internal::finite_diff_hessian_times_vector_auto()`, `stan::math::hessian_times_vector()`, and `stan::math::hessian`.
</off-topic>

src/model_rng.hpp
```diff
@@ -162,13 +162,37 @@ class bs_model {
...
 
+  /**
+   * Calculate the log density and the product of the Hessian with the specified
+   * vector for the specified unconstrain parameters and write it into the
+   * specified value pointer and Hessian-vector product pointer, dropping
+   * constants it `propto` is `true` and including the Jacobian adjustment if
+   * `jacobian` is `true`.
+   *
+   * @note If `BRIDGESTAN_AD_HESSIAN` is not defined, the complexity of this
+   * function goes from O(N^2) to O(N^3), and the accuracy of the result is
```
roualdes, line 180: Is this correct?  If `BRIDGESTAN_AD_HESSIAN` is not defined, then we're using an approximation.  The approximation should be faster with reduced accuracy, right?

roualdes (approved):
A few comments/questions.  Mostly nitpicks, I think.

Otherwise, it looks good to me.  Thanks for orchestrating yet another multiple repository feature-add.

julia/test/model_tests.jl
WardBrian: Yes, the Python interface tests run some tests with both BRIDGESTAN_AD_HESSIAN and without

src/bridgestan.h
WardBrian: I think I meant to delete that ` the`. Fixed.

src/model_rng.cpp
WardBrian: Agreed, we should open an issue in the Math repository.

src/model_rng.hpp
WardBrian: This was true when we were calculating the Hessian and then multiplying, I just missed this when I updated the other docs after the introduction of `finite_diff_hessian_times_vector_auto`

julia/test/model_tests.jl
WardBrian: I think my formatter updated this on my behalf (or at least, it's been long enough that I'd forgotten doing it)

src/bridgestan.h
bob-carpenter: perterbed -> perturbed

Can this really be done with just two gradient evals?

WardBrian: Yes, it only needs two. The best explanation is probably Justin's blog: https://justindomke.wordpress.com/2009/01/17/hessian-vector-products/

src/bridgestan.cpp
roualdes: Looks good.  I just didn't know this.  Thanks.

roualdes (approved):
Thanks for the fixes.

PR #128, opened 2023-06-13
Python: Warn if user has 5+ copies in .bridgestan

### LINKED ISSUE #83: Warn users if they have many copies of bridgestan
See https://github.com/roualdes/bridgestan/pull/80#issuecomment-1464645491 and discussion:

In both Python and Julia, a user could eventually end up with many downloaded copies due to version bumps over time. For Python, this is easy to detect as they will be in `~/.bridgestan/bridgestan-VERSION`. In Julia, it is more difficult, since they will live under `~/.julia/artifacts` as git-tree-hashes, not useful names.

WardBrian:
Closes #83

### DIFF
--- a/python/bridgestan/__init__.py
+++ b/python/bridgestan/__init__.py
@@ -6,7 +6,6 @@ __all__ = ["StanModel", "set_bridgestan_path", "compile_model"]
 
 import platform as _plt
 
-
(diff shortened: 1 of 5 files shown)


roualdes (approved):
Looks good to me.  Thanks.

PR #140, opened 2023-06-26
Add links to documentation of _R functions in C API

WardBrian:
Adds links to the standard functions for all `_R` functions in the C API.


roualdes (approved):
Looks good.  I assume you've checked that this works as intended.

PR #144, opened 2023-06-27
Delete extra sentence in R docs

WardBrian:
I believe this was a copy/paste error


bob-carpenter (approved):
thanks---definitely looks like cut-and-paste error

Issue #145, opened 2023-06-29
Julia tests require STAN_THREADS=true, doc this


roualdes:
The Julia tests require `STAN_THREADS=true` since they test threaded calls to `log_density_gradient`.  We should add a sentence in the doc about this, similar to what we did in the doc for the Rust interface, re [PR 130](https://github.com/roualdes/bridgestan/pull/130).

PR #147, opened 2023-06-29
Add to Julia interface docs required STAN_THREADS for tests

roualdes:
This PR addresses issue [#145](https://github.com/roualdes/bridgestan/issues/145) by adding a section to the Julia interface docs detailing the multi-threaded capabilities of Julia, how they coordinate with the BridgeStan API, and the fact that we test against this, thus requiring `STAN_THREADS=true` to be set for our Julia tests to run successfully.

I'd prefer to link to the function `compile_model`, but I don't know how to set up an internal link to that specific anchor in our docs.

### DIFF
--- a/julia/docs/src/julia.md
+++ b/julia/docs/src/julia.md
@@ -66,6 +66,14 @@ An example program is provided alongside the Julia interface code in `example.jl
 </details>
 ```
 
+## Tests
+
+Julia's multi-threading capabilities allow different processors/threads to make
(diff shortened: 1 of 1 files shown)


WardBrian:
How would you feel about putting this in CONTRIBUTING.md rather than the user docs?

roualdes:
Ya, that makes sense, since this is mostly about testing and we already mention in the customizing bridgestan section of Getting Started that one needs stan_threads set for parallel calls of a stan program.

Give me a second and I'll make the change.

PR #149, opened 2023-06-30
Fix: Windows TBB path logic fails if BridgeStan is not yet downloaded

### LINKED ISSUE #148: TBB path issue on Windows
Since the last release my CI tests for nutpie fail on windows. If my remote debugging is right, this is due to the changes introduced in #118. This adds some code that is executed in `__init__`, ie right at import time, where it calls `os.add_dll_directory` with contents of the `~/.bridgestan` directory. But if this runs on a fresh install, where stan was not downloaded already, this fails with a FileNotFound error.

Some logs from the CI:

https://github.com/pymc-devs/nutpie/actions/runs/5419095619/jobs/9851843966#step:8:335

If my interpretation of this is correct, this could be fixed by delaying the call to `_windows_path_setup` until after bridgestan is downloaded, ie by setting a flag to False at import time, and checking that flag each time a model is compiled. (strictly speaking it would be enough to check when a shared lib is loaded, but if, say, someone would want to load the library using a different interface, that interface would then have to do that...)

WardBrian:
Closes #148 

This also adds a very trivial test: Can the package be imported without `BRIDGESTAN` being set ahead of time.


WardBrian:
I'd like to publish a 2.1.1 shortly after this, assuming nothing new comes in

aseyboldt:
Seems I can't easily test this PR directly in the nutpie CI, because checking out stan fails with a "path too long" error...
https://github.com/pymc-devs/nutpie/actions/runs/5424318833/jobs/9863510219?pr=42#step:8:32

But this looks good to me, assuming tests pass.

WardBrian:
> fails with a "path too long" error...

Oh man, I recently dealt with this for Facebook's prophet package. Pip seems to like really long base path names, and a lot of the bash-on-Windows tools (including git, but also things like `mingw32-make`), do not respect the LONG_PATHS setting and fail if the path is over ~280 characters. Super annoying

You may be able to get things going by setting `%TEMPDIR%` to somewhere in the top-level of `C:\`, but that might still be too long. Git cloning it somewhere and then `pip install -e .` in that location would probably work.

WardBrian:
@roualdes thoughts on this fix and a .1 release?

roualdes (approved):
Ya, this all seems reasonable to me.  Is it possible for @aseyboldt to test this once it is accepted, but before we do a .1 release?

aseyboldt:
This does indeed fix the issue.
I had to call the setup function manually though, loading the library failed without it.

WardBrian:
Calling the setup function manually seems like a reasonable requirement for users who intend to load the dll not using the Python class we provide

Issue #151, opened 2023-07-05
Twice the compile time compared to cmdstan


aseyboldt:
If I compile models using cmdstan, it seems to consistently finish about twice as fast as if I compile a model with bridgestan:

```stan
// justatest.stan
parameters {
    real x;
}
model {
    x ~ normal(0, 1);
}
```

Then 
```
%%time
compiled = bridgestan.compile_model("justatest.stan")
# Wall time: 15.1 s

%%time
model = cmdstanpy.CmdStanModel("foo", "justatest.stan")
# Wall time: 7.18 s
```

I would have expected pretty much no difference, or even a bit faster comiles for bridgestan, because it doesn't have to also compile the sampler and the command line tool?

bob-carpenter:
To test compilation time, I'd put copies of the model in different places to make sure CmdStan isn't just reusing the BridgeStan compiled model.   Or try to clean out the compiled versions in between.  But if that's the difference, I'd expect CmdStanModel to be a split second.

In CmdStan, we only have to translate the Stan program to a C++ class.  The sampler and command-line tools are compiled in separate translation units and then linked.  The linking does need to happen, though, so that takes a bit of time.

Do you know if BridgeStan and CmdStanPy compile at the same optimization level?  I hope CmdStanPy isn't default to `O = 0`, as that would be very slow.  @mitzimorris should know.

WardBrian:
I wonder if it’s slower to compile position independent code? It’s also possible some of the precompiled header logic in CmdStan didn’t make it into our makefile, though that stuff is very brittle in my experience

aseyboldt:
I don't think the difference is due to caching, just to make sure I gave it slightly different models, and that didn't change the timings. I'm also seeing the pretty consistently in by benchmarks with posteriordb.

> In CmdStan, we only have to translate the Stan program to a C++ class. The sampler and command-line tools are compiled in separate translation units and then linked. The linking does need to happen, though, so that takes a bit of time.

Ah, that does make sense. Still doesn't explain the difference I think though...

> Do you know if BridgeStan and CmdStanPy compile at the same optimization level? I hope CmdStanPy isn't default to O = 0, as that would be very slow. @mitzimorris should know.

I don't see a big difference in time for a gradient evaluation between bridgestan and cmdstan, so I don't think that's it either. A small difference might be lost in all the other things that differ, but I'm pretty sure something like `O0` would be very noticeable.

[...]

PR #152, opened 2023-07-06
Fix Sundials `make` config

aseyboldt:
Previously there was no test that needed correct linking of sundials (the ode test was only using the Runge Kutta solver).
This fails locally for me, but I would like to see if it also does so in the CI, or if this is just an issue on my machine.

It is possible that the makefile config that leads to incorrect sunidals linking is also responsible for the compile times in #151.

Update: It also fails in the CI. Here it looks like the linux tests might pass, but locally they fails for me when the module is loaded (for instance in the python tests).


WardBrian:
This was caused by https://github.com/stan-dev/math/pull/2885#pullrequestreview-1356896247, the fix is to change LIBSUNDIALS to SUNDIALS_TARGETS in our makefile

aseyboldt:
That does seem to fix the issue :-)
I had to manually set `CXXFLAGS_SUNDIALS += -fPIC`, is it expected that CXXFLAGS is not used in the sundials build commands?

WardBrian:
Let me look into CXXFLAGS. We inherit all that from the stan-dev/math makefiles, and they can be a bit opaque.

I think we can delete the existing RK45 ode test - I didn't check that the solver was actually from sundials, which is my bad.

PR #153, opened 2023-07-06
fix(rust): Allow loading more recent libraries

aseyboldt:
I think it should be legal to provide a bridgestan library that was compiled with a later version of bridgestan, if the major version still matches. But the rust wrapper expects an exact version match right now. This means that downstream rust libraries have to specify an exact version of the python bridgestan library, or a user could update the python library and libraries built by that new bridgestan couldn't be loaded anymore.

Supporting *older* libraries would be some work, as we would have to handle the case that symbols might be missing in the library, so we still don't allow that.

This change would also imply that the ABI can't change without a major version bump. I can't really think of reasons why that would be a problem though...


WardBrian:
> This change would also imply that the ABI can't change without a major version bump.

We only really guarantee a stable API, but ABI compatibility should be inherited from whatever C compiler is used. This sounds good to me

PR #157, opened 2023-07-12
Clarify usage of `set_bridgestan_path` in examples

### LINKED ISSUE #156: Better Error Message for Python Interface Automatically Downloading BridgeStan Source
BridgeStan's Python interface can automatically download the BridgeStan source code "the first time you need it", as the documentation describes. In the code, this is performed by `get_bridgestan_src()` which is called by `get_bridgestan_path` which is called by `compile_model()`. Therefore "the first time you need" BridgeStan source code really means the first time a model is compiled.

While this makes sense, most of the (limited) BridgeStan examples I've seen begin with the `set_bridgestan_path()` function.  When using BridgeStan for the first time, `set_bridgestan_path()` will call `verify_bridgestan_path()` and throw an error (since BridgeStan source code is not installed and `get_bridgestan_src()` was never called).

To resolve this issue, I had to dive into the source code and call `get_bridgestan_src()` manually. If indeed most BridgeStan projects begin with `set_bridgestan_path()`, I imagine many other new users may face the same issue.

I want to suggest two small improvements to fix this:
1. In the documentation, replace "the first time you need it" with "the first time a model is compiled"
2. Update the error message in `verify_bridgestan_path()` to suggest to users how to download the BridgeStan source code. Perhaps something like: "if this is your first time using BridgeStan, call `get_bridgestan_src()` immediately before the `set_bridgestan_path()` function." This may, however, negate the automatic downloading of BridgeStan source code that was intended in the first place.

Any suggestions or feedback is welcome.

WardBrian:
Closes #156

### DIFF
--- a/docs/getting-started.rst
+++ b/docs/getting-started.rst
@@ -24,7 +24,8 @@ and no additional dependencies are required to be installed separately for the C
 
 As of version 1.0.2, the :doc:`Julia <languages/julia>` and
 :doc:`Python <languages/python>` interfaces will download
-the source for you, so this section is optional.
+the source for you the first time you compile a model, so
+this section is optional.
 
 Downloading a released archive
 ______________________________
--- a/docs/languages/julia.md
+++ b/docs/languages/julia.md
@@ -37,7 +37,7 @@ BridgeStan is registered on JuliaRegistries each release.
 ```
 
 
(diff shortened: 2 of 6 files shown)


roualdes:
I agree this is a doc issue.  The updated language makes sense to me, but I'm not really the intended audience.  @gil2rok, what do you think of these additional comments in the doc and the examples?  Do you imagine these changes would have helped you understand what we have in mind for the use case of `set_bridgestan_path()`?

gil2rok:
Yes, this type of comment would have helped clarify when / how to use `set_bridgestan_path()`. Thanks!

PR #159, opened 2023-07-19
JOSS: Add self as author and mention rust interface

aseyboldt:


paper/paper.md
```diff
@@ -25,13 +26,18 @@ authors:
   - name: Seth D. Axen
     orcid: 0000-0003-3933-8247
     affiliation: 3
+  - name: Adrian Seyboldt
```
sethaxen, line 29: Since your contribution was more extensive than mine, I'd suggest you be placed before me in the author list.

aseyboldt: Got to admit I'm pretty uncertain with all the conventions in publications. Thanks :-)

bob-carpenter:
I don't mind one way or another.  Since @aseyboldt was asking, let me lay out the standard conventions.  The "first authors" on the list do the lion's share of the research and/or writing.  The "middle authors" make contributions worthy of going on the paper.  The "last authors" are the lab directors/supervisors.  In enormous physics or bio lab projects, there can be a dozen first authors and hundreds of middle authors.  The very first author gets a boost in name recognition because the paper will often be cited as (First Author et al. year) in long citation format.  The most senior author usually gets a boost in name recognition because they're usually the best known author and thus easier for people to remember.

In bio, they often include a description of what each author contributed.  I really liked the way the transformers paper just randomized the author order and then included a short note on what each author contributed.

PR #165, opened 2023-08-10
Remove unnecessary backward pass in log_density

WardBrian:
In `bs_log_density`, if you request dropping of constant terms, it is important we instantiate the model class' log_prob implementation with `var`s (`double`s are considered constants, and therefore dropped).

However, it is _not_ necessary to actually compute the backward pass on those vars, which is currently done in log_density when `propto=True`. This means the log density calculation is slower than it could be, which this PR fixes.


bob-carpenter (approved):
good catch and the fix looks good

PR #171, opened 2023-09-21
Update CI and Makefile

WardBrian:
This fixes CI failures caused by the recent [changes to Github Actions Runners](https://github.com/actions/runner-images/releases/tag/win22%2F20230918.1) to use Mingw 12.2.0

Something like this will hopefully be incorporated into Stan's makefiles in the future.


roualdes (approved):
Great.  Thanks for keeping us one step ahead.

PR #177, opened 2023-10-04
Move all python metadata to pyproject.toml

WardBrian:
See [PEP 621](https://peps.python.org/pep-0621/). This consolidates the information and removes the need for us to run any code at install time to determine our own version.


python/pyproject.toml
```diff
@@ -1,3 +1,34 @@
...
+    { name = "Brian Ward", email = "bward@flatironinstitute.org" },
+    { name = "Bob Carpenter", email = "bcarpenter@flatironinstitute.org" },
+    { name = "Edward Roualdes", email = "eroualdes@csuchico.edu" },
+]
+requires-python = ">=3.9"
+dependencies = ["numpy"]
+dynamic = ["version"]
+
+[tool.setuptools.dynamic]
+version = { attr = "bridgestan.__version.__version__" }
```
roualdes, line 19: So this line is `eval()`ed or something like it at install?

roualdes (approved):
More simplifying.  Love it.  

Just one question for my understanding, but otherwise looks good to me.  Thanks!

python/pyproject.toml
WardBrian: In a limited way. It can either read a module attribute (like this) or a text file: https://setuptools.pypa.io/en/latest/userguide/pyproject_config.html#dynamic-metadata

PR #178, opened 2023-10-13
Python: Issue warning when re-loading the same module

WardBrian:
@charlesm93 recently ran into the issue where re-loading the same shared library is a no-op on some systems. 

We warn against this in Julia and R (#45), but in Python the standard library was missing a way to list all already-loaded dlls, so it was left out.

After Charles' prompting, I wrote one: https://github.com/WardBrian/dllist
This small Python package essentially re-implements the same platform-specific logic that the Julia `dllist` function that we use does. I finished it up this morning and put it on PyPI, so if we don't mind a small extra dependency we can add this warning now for Python users as well.

### DIFF
--- a/.github/workflows/main.yaml
+++ b/.github/workflows/main.yaml
@@ -115,8 +115,8 @@ jobs:
 
       - name: Run tests
         run: |
+          pytest -v python/
           cd python/
-          pytest -v
           python example.py
         env:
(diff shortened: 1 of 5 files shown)


roualdes (approved):
This is great.  Indeed, that is a small dependency.  Ok by me.

PR #180, opened 2023-10-24
Fix memory leak in log_density(propto=True)

WardBrian:
When I submitted #165 to fix the unnecessary computation in `log_density(propto=True)` I didn't consider that removing the call to `grad` also meant the `var` stack was no longer being cleared, so the memory would build up over time.

You can see this pretty easily in the current version, e.g.:

```python
import numpy as np
import bridgestan

m = bridgestan.StanModel("./test_models/gaussian/gaussian_model.so", "./test_models/gaussian/gaussian.data.json")

def print_lp():
    p = np.random.randn(2)
    lp = m.log_density(p, propto=True)
    print(lp)

for _ in range(100_000):
    print_lp()
```

Running this in a repl, you can see that the memory used by the process grows throughout the loop and stays high. After this PR, it does not.

### DIFF
--- a/src/model_rng.cpp
+++ b/src/model_rng.cpp
@@ -263,6 +263,7 @@ void bs_model::log_density(bool propto, bool jacobian, const double* theta_unc,
     } else {
       *val = model_->log_prob_propto(params_unc_var, outstream).val();
     }
+    stan::math::recover_memory();
   } else {
     if (jacobian) {
       *val = model_->log_prob_jacobian(params_unc, outstream);
--- a/src/model_rng.cpp
+++ b/src/model_rng.cpp
@@ -256,13 +256,21 @@ void bs_model::log_density(bool propto, bool jacobian, const double* theta_unc,
   if (propto) {
     // need to have vars, otherwise the result is 0 since everything is
     // treated as a constant
-    Eigen::Matrix<stan::math::var, Eigen::Dynamic, 1> params_unc_var(
-        params_unc);
-    if (jacobian) {
-      *val = model_->log_prob_propto_jacobian(params_unc_var, outstream).val();
-    } else {
-      *val = model_->log_prob_propto(params_unc_var, outstream).val();
+    try {
+      Eigen::Matrix<stan::math::var, Eigen::Dynamic, 1> params_unc_var(
+          params_unc);
+      if (jacobian) {
+        *val
+            = model_->log_prob_propto_jacobian(params_unc_var, outstream).val();
+      } else {
+        *val = model_->log_prob_propto(params_unc_var, outstream).val();
+      }
+    } catch (...) {
+      // because we created vars on the stack, we need to recover memory
+      stan::math::recover_memory();
+      throw;  // re-caught by top level exception logic
     }
+    // also recover memory if no exception was thrown
     stan::math::recover_memory();
   } else {
     if (jacobian) {


bob-carpenter (changes requested):
Thanks.  I should have caught that the first time.

I don't think this is the right pattern.  I think you need the C++ equivalent of Java's `finally`.  Namely:

```cpp
try {
  ... everything ...
} catch (...) {
   stan::math::recover_memory():
   return;
}
stan::math::recover_memory();
```

Now that I write that, maybe only the second recover is necessary. 

It'd be a bonus to convert the C++ exceptions into Python exceptions on the other side.

WardBrian:
Good point. We do catch the top level exceptions somewhere else, so we can catch, recover memory, and re-raise

WardBrian:
`...` includes `std::exception`, so there's no need to separate the cases here (they are separate in the top-level function that calls this, though!). We have a test for the log density failing with an error message which is still passing

src/model_rng.cpp
```diff
@@ -256,13 +256,22 @@ void bs_model::log_density(bool propto, bool jacobian, const double* theta_unc,
...
+      if (jacobian) {
+        *val
+            = model_->log_prob_propto_jacobian(params_unc_var, outstream).val();
+      } else {
+        *val = model_->log_prob_propto(params_unc_var, outstream).val();
+      }
+    } catch (...) {
+      // because we created vars on the stack, we need to recover memory
+      stan::math::recover_memory();
+      throw;  // re-caught by top level exception logic
```
bob-carpenter, line 271: Cool!  I didn't know or remember you cold rethrow.  I really don't think this needs any doc as it's a standard pattern, but I'll leave that up to you.  I generally don't like comments that say what callers are going to do with the results as it's so open-ended.

src/model_rng.cpp
```diff
@@ -256,13 +256,22 @@ void bs_model::log_density(bool propto, bool jacobian, const double* theta_unc,
...
     }
+    // also recover memory if no exception was thrown
+    stan::math::recover_memory();
   } else {
     if (jacobian) {
```
bob-carpenter, line 276: The try/catch needs to be one level up to handle the else case, too.  Anything that calls log prob and implicates autodiff.  I think these do.  But let me know if there's a reason the next calls won't use autodiff stack.

bob-carpenter (changes requested):
Thanks.  I think it needs to go one level up, so if you could please check that, I'd appreciate it.

src/model_rng.cpp
WardBrian: params_unc is just a vector of doubles, so there won’t be any autodiff in the overload of log_prob it ends up calling.

I actually used the propto=False case to confirm that the memory leak was originally specific to this branch of things, so it should be good to go

src/model_rng.cpp
WardBrian: I would generally agree, but since this class isn’t part of our public API and we are its only callers I feel fine saying so

WardBrian:
The other option we could take is using what `gradient` itself uses which is the [`nested_rev_autodiff`](https://github.com/stan-dev/math/blob/develop/stan/math/rev/core/nested_rev_autodiff.hpp) RAII class which cleans up memory in its destructor. We don't need nested autodiff (since we don't need _any_ autodiff) but I believe it would be functionally equivalent in this case, and perhaps a bit cleaner.

bob-carpenter (approved):

bob-carpenter:
Thanks.  Happy to review another PR if you think it'll be cleaner.

PR #184, opened 2023-11-06
Improve developer documentation and experience

WardBrian:
- Cleans up CONTRIBUTING.md to push even more information into the online/rendered documentation
- Improves the makefile error provided when the submodules are missing (similar to https://github.com/stan-dev/cmdstan/pull/1217)
- Added a makefile target to format all code
- Made the `.pylintrc` in `python/` a bit more useful for our project (we still don't enforce this anywhere)

### DIFF
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -2,7 +2,7 @@
 
 We welcome contributions to the project in any form, including bug reports, bug fixes, new features, improved documentation, or improved test coverage.
 
-**Developer-specific documentation is available at [as part of our online documentation](https://roualdes.github.io/bridgestan/latest/internals.html)**. Information on building BridgeStan, running tests, and developing with the project is available there.
+**Developer-specific documentation is available [as part of our online documentation](https://roualdes.github.io/bridgestan/latest/internals.html)**. Information on building BridgeStan, running tests, and developing with the project is available there.
 
 This document houses additional information about licensing and procedures for contributing to the project.
 
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -2,7 +2,9 @@
 
 We welcome contributions to the project in any form, including bug reports, bug fixes, new features, improved documentation, or improved test coverage.
 
-Developer-specific documentation is available at https://roualdes.github.io/bridgestan/latest/internals.html
+**Developer-specific documentation is available at [as part of our online documentation](https://roualdes.github.io/bridgestan/latest/internals.html)**. Information on building BridgeStan, running tests, and developing with the project is available there.
+
+This document houses additional information about licensing and procedures for contributing to the project.
 
 ## Licensing
 
@@ -16,7 +18,7 @@ We follow the practices laid out in the [GitHub Terms of Service](https://docs.g
 * Developers (or their assignees) retain copyright for their doc and code contributions.
 * Developers agree to release their contributed code and documentation under the repository licenses (see above).
 
-## Developers
+## Development Process
 
 We follow standard open source and GitHub practices:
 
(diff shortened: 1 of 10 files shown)


CONTRIBUTING.md
```diff
@@ -2,7 +2,9 @@
 
 We welcome contributions to the project in any form, including bug reports, bug fixes, new features, improved documentation, or improved test coverage.
 
-Developer-specific documentation is available at https://roualdes.github.io/bridgestan/latest/internals.html
+**Developer-specific documentation is available at [as part of our online documentation](https://roualdes.github.io/bridgestan/latest/internals.html)**. Information on building BridgeStan, running tests, and developing with the project is available there.
```
roualdes, line 5: strike at

Makefile
```diff
@@ -141,3 +157,19 @@ $(STANC):
...
+# We assume that means the submodule is missing
+##
+$(MATH)make/% :
+	@echo 'ERROR: Missing Stan submodules.'
+	@echo 'We tried to find the Stan Math submodule at:'
+	@echo '  $(MATH)'
+	@echo ''
+	@echo 'The most likely source of the problem is BridgeStan was cloned without'
+	@echo 'the --recursive flag.  To fix this, run the following command:'
+	@echo '  git submodule update --init --recursive'
```
roualdes, line 172: Is there anway, maybe via extra bits in our top level `.gitmodules` file, to include `--depth=1`, `--recurse-submodules`, and `--shallow-submodules`?  I read the man page on `git-submodule`, but didn't come away with an obvious answer.

roualdes (approved):
One suggestion, one question, and otherwise looks good.  Thanks.

Makefile
WardBrian: You can put `shallow = true` in there but it sounds like that [may be buggy](https://stackoverflow.com/a/47374702)

roualdes: Ok.  Thanks for double checking.

roualdes (approved):

CONTRIBUTING.md
bob-carpenter: I prefer links to be minimal nouns or noun phrases, so I would use `as part of our [online documentation](...)`, though it could be `as part of [our online documentation](...)` or even `as [part of our online documentation](...)`.

CONTRIBUTING.md
```diff
@@ -2,7 +2,9 @@
...
+
+This document houses additional information about licensing and procedures for contributing to the project.
```
bob-carpenter, line 7: houses -> contains OR supplies OR provides, because there's no housing, so it's distracting.

docs/internals/development.rst
```diff
@@ -0,0 +1,55 @@
...
+Building BridgeStan
+-------------------
+
+Developers should follow the instructions in :doc:`../getting-started` to ensure
+they have the minimal C++ build tools required for working with BridgeStan.
+
+C++
+---
+
+* We use the C++1y standard for compilation (``-std=c++1y`` in both clang and gcc). This is partway between C++11 and C++14, and is what Stan requires.
```
bob-carpenter, line 18: As of R 4.3, they support C++17.  It rolled out earlier this year with a late 4.2.x.  See https://developer.r-project.org/blosxom.cgi/R-devel/NEWS/2023/01/27#n2023-01-27

I don't know if Python is a blocker.

docs/internals/development.rst
```diff
@@ -0,0 +1,55 @@
...
+
+* We try to write standards-compliant code that does not depend on features of specific platforms (except where needed for compatibility).
+  Specifically, we do not use OS-dependent or compiler-dependent C++.
+  Our C++ code does not depend on the ``R.h`` or ``Python.h`` headers, for example.
+  On the other hand, adding new signatures to work with a specific language's style of foreign function interface is permitted
+  (an example can be found in the :ref:`R compatibility functions <R-compat>`, which requires a particular pointer-based style).
+
+* We try to follow the `Google C++ Style Guide <https://google.github.io/styleguide/cppguide.html>`_, but (a) we allow C++ exceptions, and (b) we allow reference arguments.
+
+* We recommend using `Clang format <https://clang.llvm.org/docs/ClangFormat.html>`_ with our config file `.clang-format <https://github.com/roualdes/bridgestan/blob/main/.clang-format>`_.
```
bob-carpenter, line 28: I think this line's too long.  I think it's OK to wrap even with bullets.

docs/internals/development.rst
```diff
@@ -0,0 +1,55 @@
...
+
+* We use the C++1y standard for compilation (``-std=c++1y`` in both clang and gcc). This is partway between C++11 and C++14, and is what Stan requires.
+
+* We try to write standards-compliant code that does not depend on features of specific platforms (except where needed for compatibility).
+  Specifically, we do not use OS-dependent or compiler-dependent C++.
+  Our C++ code does not depend on the ``R.h`` or ``Python.h`` headers, for example.
+  On the other hand, adding new signatures to work with a specific language's style of foreign function interface is permitted
+  (an example can be found in the :ref:`R compatibility functions <R-compat>`, which requires a particular pointer-based style).
+
+* We try to follow the `Google C++ Style Guide <https://google.github.io/styleguide/cppguide.html>`_, but (a) we allow C++ exceptions, and (b) we allow reference arguments.
```
bob-carpenter, line 26: There seems to be a dangling underscore after the URL.   Or is that just indicating some kind of space?  We definitely don't want an underscore but I also don't think we want a space between a formatted URL and punctuation.

bob-carpenter (changes requested):
Just a few comments on doc.  I like all the changes to Python that move things up out of else conditions when the main if just fires off an exception that will reroute control.

docs/internals/development.rst
WardBrian: Because we don't rely on any specific C++ ABI, we could use any C++ standard we like for this project. This comment is just there because it's what our makefiles pass and what the rest of Stan uses

docs/internals/development.rst
WardBrian: That's just how links look in [RST](https://www.sphinx-doc.org/en/master/usage/restructuredtext/basics.html#hyperlinks)

bob-carpenter: I looked it up and wow, they really do include the underscore to mean it's an inline link.  So I retract all those comments about not wanting spaces---I had thought it was just GitHub highlighting a space :-).

bob-carpenter:
Sorry--I didn't see @roualdes approving this or I wouldn't have added other picky comments (though I would still prefer to see the syntax of what's included in links be a noun).

PR #185, opened 2023-11-09
Set up visibility controls for public symbols in the library

WardBrian:
This PR makes the necessary changes to set the default symbol visibility to "hidden". It's worth reading more about this on the [GCC Wiki](https://gcc.gnu.org/wiki/Visibility).

The summary:
- By default, every symbol in a shared library is available to the linker
- If you are explicit about which symbols _actually should be public_, your compiler can then perform optimizations based on the assumption that the remaining code doesn't need to be accessible
- The Microsoft compilers do this by default, with the added wrinkle of also distinguishing between "public and provided by the current translation unit" (`dllexport`) vs "public and provided by someone else" (`dllimport`)

### DIFF
--- a/.github/workflows/main.yaml
+++ b/.github/workflows/main.yaml
@@ -8,7 +8,7 @@ on:
   workflow_dispatch: {}
 
 env:
-  CACHE_VERSION: 0
+  CACHE_VERSION: 1
 
 # only run one copy per PR
 concurrency:
--- a/Makefile
+++ b/Makefile
@@ -18,8 +18,9 @@ include $(MATH)make/compiler_flags
 include $(MATH)make/libraries
 
 ## Set -fPIC globally since we're always building a shared library
-CXXFLAGS += -fPIC
+CXXFLAGS += -fPIC -fvisibility=hidden -fvisibility-inlines-hidden
 CXXFLAGS_SUNDIALS += -fPIC
+CPPFLAGS += -DBRIDGESTAN_EXPORT
 
 ## set flags for stanc compiler (math calls MIGHT? set STAN_OPENCL)
 ifdef STAN_OPENCL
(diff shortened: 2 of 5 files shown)


roualdes (approved):
Not sure how you found such a thing, but thanks for sharing the reading material about it.  Looks good.  Just a couple of questions.

1. What is the use case of BRIDGESTAN_EXPORT?  It doesn't seem like we use it, or maybe I just missed that.  Should this have some doc about it?
2. Have you done any speed or size measurements based on this change?  I'm happy to move forward without it, but generally curious.  The linked reading material suggests big gains for a big project and I don't think BridgeStan is very large.

WardBrian:
> What is the use case of BRIDGESTAN_EXPORT?  It doesn't seem like we use it, or maybe I just missed that.

It gets defined on line 23 of the makefile. The purpose is such that when we're building a shared library on Windows, the functions are marked as `dllexport`, but if _somebody else_ writes `#include <bridgestan.h>`, they get a `dllimport`-marked function instead.

> Should this have some doc about it?

It could, but I'm not sure what to write or where it would be best suited. Is the idea we'd be documenting it for end users or for developers/ourselves?

> size measurements

Binary size (using model `ode_sundials`):

| Optization setting | Size in bytes, before | Size in bytes, after |
|--------|--------|--------|
| `O3` | 1,108,800 | 1,016,888 |
| `O2` | 1,065,152 | 969,376 |
| `Os` | 1,041,056 | 860,928 |

Number of symbols (`nm -C -D ode_sundials_model.so | wc -l`). These are both at O3, it mattered much less for this in terms of total numbers:

| Before | After |
|--------|--------|
| 1559 | 956 |

Note: I picked this model as sort of a worst-case, since the sundials symbols are exported. For something like bernoulli, the "after" is ~500 symbols.

> speed

This is harder since we don't have a ton of precise benchmarks. At a basic level, things like loading the library will get faster (less to load into memory, and less symbols to scan over to find our functions), but testing to see if this genuinely provides an increased opportunity for optimizations is difficult

roualdes:
Re BRIDGESTAN_EXPORT.  Thanks. That makes sense.

This seems great.  Thanks again for finding this and thanks for incorporating it into BridgeStan.

PR #188, opened 2023-11-29
Unify constructors to always allow a .stan file to be passed

### LINKED ISSUE #187: More consistent instantiation from a .stan file in various languages
The changes I'm proposing in #172 do not expose a separate constructor for .stan files compared to .so files, but rather just branch on the extension of the file passed as the first argument.

So, in R, it would look like:

```R
StanModel$new("foo_model.so", data="foo.data.json") # loads a model already compiled
StanModel$new("bar.stan", data="bar.data.json") # compiles, *then* loads
```

In Python this looks different for the two cases:

```python
StanModel("foo_model.so", data="foo.data.json") # loads a model already compiled
StanModel.from_stan_file("bar.stan", data="bar.data.json") # compiles, *then* loads
```

In Julia it looks more like R, but is technically doing type dispatching rather than checking the .stan extension:

```julia
StanModel("foo_model.so", data="foo.data.json") # loads a model already compiled
StanModel(stan_file="bar.stan", data="bar.data.json") # compiles, *then* loads
```

I think the real outlier here is Python, which deserves to be unified IMO. We can leave the existing function, but just update some logic in the constructor to handle .stan files. Whether Julia should also be updated or left using this disbatching I am less sure.

WardBrian:
Closes #187.

Changes:
- Python: `StanModel.from_stan_file` is deprecated. The logic is subsumed in the constructor, which will trigger compilation if the file provided ends in `.stan`
- Python: The `model_data` argument is renamed to `data`. The previous argument still works but issues a warning.
- Python: All functions which accepted strings for file paths now also accept the [`os.PathLike`](https://docs.python.org/3/library/os.html#os.PathLike) interface
- Julia: The `StanModel(;stan_file,...)` overload is deprecated. This logic is now subsumed in the constructor, which will trigger compilation if the file provided ends in `.stan`

### DIFF
--- a/python/bridgestan/model.py
+++ b/python/bridgestan/model.py
@@ -1,7 +1,8 @@
 import ctypes
 import warnings
+from os import PathLike, fspath
 from pathlib import Path
-from typing import List, Optional, Tuple
+from typing import List, Optional, Tuple, Union
 
 import numpy as np
 import numpy.typing as npt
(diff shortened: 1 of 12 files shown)


python/bridgestan/model.py
```diff
@@ -32,22 +33,33 @@ class StanModel:
 
     def __init__(
         self,
-        model_lib: str,
-        model_data: Optional[str] = None,
+        model_lib: Union[str, PathLike],
+        data: Union[str, PathLike, None] = None,
```
roualdes, line 37: Is `None` better here than `""`?

python/bridgestan/model.py
```diff
@@ -32,22 +33,33 @@ class StanModel:
...
     ) -> None:
         """
         Construct a StanModel object for a compiled Stan model and data given
         constructor arguments.
 
-        :param model_lib: A system path to compiled shared object.
-        :param model_data: Either a JSON string literal, a
+        :param model_lib: A system path to compiled shared object or a ``.stan``
+            file to be compiled.
+        :param data: Either a JSON string literal, a
```
roualdes, line 51: Would it be helpful to have listed `:param model_data:` also, with a note about it being deprecated in favor of `data`?

python/bridgestan/model.py
```diff
@@ -685,6 +674,58 @@ def _handle_error(self, err: ctypes.c_char_p, method: str) -> Exception:
 
         return RuntimeError(f"Unknown error in {method}. ")
 
+    @classmethod
```
roualdes, line 677: I'm still on the page of putting a deprecation warning in the function doc.  What do you think of this idea in general?  I'll stop pointing it out for the rest of this review.

roualdes (approved):
Overall this looks great.  Just a general questions about putting deprecation warnings in doc strings, to compliment error messages.

python/bridgestan/model.py
WardBrian: None ends up being passed as `nullptr` to C++, which then creates an `empty_var_context`:

https://github.com/roualdes/bridgestan/blob/41b66b7956817ecb411e7dccedeed4d6fd8347dc/src/model_rng.cpp#L65C1-L67C56

If you pass `""` instead, this string is then treated as an empty JSON, which creates a json parser/var_context and populates it with nothing. 

So they're semantically equivalent, but trigger different code paths

python/bridgestan/model.py
WardBrian: I tend to not list renamed arguments in docstrings, but I'm happy to do so if you'd rather

Issue #190, opened 2023-12-03
Python callback into BridgeStan with C types


roualdes:
I have an implementation of Stan that iterates and does warmup calculations in Python and leapfrog/trajectories in C++.  The gradient calculations implicit in each leapfrog step happen via a Python callback which then calls BridgeStan: Python -> C++ -> Python/BridgeStan -> C/C++/BridgeStan.

Since the C++ code is calling a BridgeStan function, and Python/BridgeStan expects numpy arrays, I've got to convert a `double*` to a numpy array just so that BridgeStan can convert the numpy array back to `double*`.  The conversion to a numpy array is unnecessarily costly, especially since everything is using the same underlying memory.

Consider BridgeStan's [python/bridgestan/model.py lines 219 to 229](https://github.com/roualdes/bridgestan/blob/4d2ee05d6a4be57c07247706e2a60581750db481/python/bridgestan/model.py#L219).  This code chunk defines the Python/BridgeStan interface to the C/BridgeStan function `bs_log_density_gradient`.  Lines 225 and 227 specify numpy arrays as argument types from Python/BridgeStan -> C/C++/BridgeStan.

Simply adding the following code to `python/bridgestan/model.py::StanModel`'s `__init__()` method allows a Python callback from C++ to re-use the original `double*`:

```
self._ldg = self.stanlib["bs_log_density_gradient"]
self._ldg.restype = ctypes.c_int
self._ldg.argtypes = [
    ctypes.c_void_p,
    ctypes.c_int,
    ctypes.c_int,
    ctypes.POINTER(ctypes.c_double), # changed from double_array
    ctypes.POINTER(ctypes.c_double),
    ctypes.POINTER(ctypes.c_double), # changed from double_array
    star_star_char,
]
```

What do you think of building such functionality into BridgeStan?  Just adding the above code won't break our API in anyway, but does offer a whole new world of possibility for BridgeStan.  I'm not sold on the name `_ldg`, I'm just asking about the possibility of including such functionality.

If y'all are into this idea, would you brainstorm with me what a better design might look like that allow callbacks, from the various higher level languages, into BridgeStan using plain C types?

bob-carpenter:
> What do you think of building such functionality into BridgeStan? 

I couldn't quite follow all the callbacks, so I'm not entirely sure what you are suggesting adding to BridgeStan and what that would enable for clients that they can't achieve now.

WardBrian:
[...]

PR #191, opened 2023-12-05
Misc. cleanups

WardBrian:
While looking into #190, I found a few small improvements:

- Using `ctypes.byref` is cheaper than actually constructing a `ctypes.pointer` 
- Some missing return type annotations
- A few places where we said 'Bridgestan' instead of 'BridgeStan'


roualdes (approved):
Great find.  Thanks.  

Here's a quick reference to that claim, for whatever it's worth.  Python [ctypes.byref() doc](https://docs.python.org/3/library/ctypes.html#ctypes.byref).

PR #192, opened 2023-12-05
Add implementation details docs page, note on propto=True for log_density

WardBrian:
This is a bit of a weird thing I've been thinking about recently that it is good to alert users for. It's related to #165 and #180.

Basically `propto=True` requires we pass `vars` to Stan. Before #165, we were even calling `grad`, wasting a lot of computation. Since #165, we no longer call grad, but it still may lead to more work than you'd expect, since the Stan math library assumes (justifiably) that if you're calling a function with `var`s you will want gradients, and so it can do some pre-computation for you. Because we never call/use `grad()` in `log_density`, this is wasted effort.

The big offenders are the higher order functions like `reduce_sum`, which basically calculate their entire gradients in the "forward pass".

I wrote up a docs page on implementation details and added a section about this. I'm not sure if it should be linked other places or not.

### DIFF
--- a/docs/internals/details.rst
+++ b/docs/internals/details.rst
@@ -4,31 +4,41 @@ Implementation Details
 
 .. _log_density_propto:
 
-``log_density`` with ``propto=true``
-------------------------------------
+Speed of the ``propto`` argument
+--------------------------------
 
 The log density function provided by a Stan model has
-the ability to be calculated up to an additive constant.
+the ability drop additive constants from the calculation.
 This is indicated by the ``propto`` ("``prop``\ortional  ``to``")
 argument to function.
-Usually, this is done for efficiency reasons, as the constant
-terms may require computation that is not necessary for calculating
-gradients or most sampling algorithms.
+
+If you are using an application such as an MCMC algorithm which requires
+gradients and only needs the log density up to a proportion, setting
+``propto=True`` will be at least as fast as setting ``propto=False``
+and is generally recommended (and the default value).
 
 However, in the case of the ``log_density`` function (which does not calculate
-derivatives), this argument may make the calculation **slower**. This is because
-the implementation of this argument relies on the presence of autodiff types
-(``var``\s, in the terminology of Stan's math library) to determine what is or
-is not constant with respect to the parameters. If not for this (and indeed, if
-the argument is set to ``false``), the calculation of the log density is able to be
-computed using only primitive types (``double``\s).
+derivatives), this argument has the potential to *slow down* computation, and we
+recommend setting it to ``False`` or timing it for your model of interest before
+proceeding. Note that the default value of ``propto`` is ``True`` for consistency
+with the versions of the function that do calculate gradients, so extra care is needed.
+
+Why is ``log_density`` different?
+_________________________________
+
+The implementation of the ``propto`` argument relies on the presence
+of autodiff types (``var``\s, in the terminology of Stan's math library)
+to determine what is or is not constant with respect to the parameters.
+When the argument is ``False``, the calculation of the log density is able to be
+computed using only variables of type ``double``.
 
 The consequence of this is that, if the ``propto`` argument is set to ``true``,
 the ``log_density`` function will at a minimum need to perform more allocations
 than if it were set to ``false``. There may be an even higher cost, due to functions
 such as |reduce_sum|_ or Stan's ODE integraters changing their behavior when applied
 to autodiff types and performing additional work which is thrown away when gradients
-are not needed.
+are not needed. These additional computations can quickly overwhelm any speed up
+received by dropping additive constants in practice.
 
 
 .. |reduce_sum| replace:: ``reduce_sum``
--- a/docs/internals/details.rst
+++ b/docs/internals/details.rst
@@ -1,0 +1,36 @@
+Implementation Details
+======================
+
+
+.. _log_density_propto:
+
+``log_density`` with ``propto=true``
+------------------------------------
+
+The log density function provided by a Stan model has
+the ability to be calculated up to an additive constant.
+This is indicated by the ``propto`` ("``prop``\ortional  ``to``")
+argument to function.
+Usually, this is done for efficiency reasons, as the constant
+terms may require computation that is not necessary for calculating
+gradients or most sampling algorithms.
+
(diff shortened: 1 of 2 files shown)


WardBrian:
Here is a demo of what I'm describing, using [this model](https://github.com/WardBrian/ffistan/tree/main/test_models/sir)

```python
import bridgestan
import timeit

model = bridgestan.StanModel('sir.stan', 'sir.data.json')

with open('sir.init.json') as f:
    params = model.param_unconstrain_json(f.read())


def one():
    model.log_density(params, propto=True)

def two():
    model.log_density(params, propto=False)

timeit.timeit(one, number=1000) # warms-up caches etc

time_one = timeit.timeit(one, number=1000)
time_two = timeit.timeit(two, number=1000)

print(f"propto=T: {time_one*1000:.1f}ms")
print(f"propto=F: {time_two*1000:.1f}ms")

```

This prints
```
propto=T: 329.7ms
propto=F: 81.4ms
```

So, over 4 times slower for this model!

docs/internals/details.rst
```diff
@@ -0,0 +1,36 @@
+Implementation Details
+======================
+
+
+.. _log_density_propto:
+
+``log_density`` with ``propto=true``
+------------------------------------
+
+The log density function provided by a Stan model has
```
bob-carpenter, line 10: I would phrase this as having the ability to drop constants.   Then I'd give simple recommendations:

1.  If you're running MCMC that needs gradients and only density up to proportion, then use `propto = true`.  Setting `propto=true` will be at least as fast.
2.  To evaluate the log density on `double` values to match, use `propto=true`.  Setting `propto=true` may be slower or faster, depending on the cost of calculating normalizing constants (`propto=false`) and the cost of autodiff (required to get the right answer if `propto=true`).

I don't think we need to say much more than that.

Why the double back ticks?  

I couldn't understand what lines 34/35 were doing.




I'd just give simple recommendations:

* doing gradient-based calculations with autodiff:  use propto because it's faster
* doing log density evals without autodiff:  depending

bob-carpenter (approved):
I think it's OK as is, but I added some suggestions and am happy to re-review.

docs/internals/details.rst
WardBrian: The double back ticks are how ReStructuredText (that sphinx uses) wants code formatted. They're equivalent to single backticks in Markdown. Lines 34/35 are also a RST detail to get a link that is also `code formatted`. The result is that "|reduce_sum|" above gets rendered as [`reduce_sum`](https://mc-stan.org/docs/stan-users-guide/reduce-sum.html)

I agree phrasing it in terms of a suggestion for each case is clearer. I left the explanation in, but under a sub-heading for the curious.

docs/internals/details.rst
```diff
@@ -0,0 +1,46 @@
...
+======================
+
+
+.. _log_density_propto:
+
+Speed of the ``propto`` argument
+--------------------------------
+
+The log density function provided by a Stan model has
+the ability drop additive constants from the calculation.
```
roualdes, line 11: "the ability drop" -> "the ability to drop"

docs/internals/details.rst
```diff
@@ -0,0 +1,46 @@
...
+This is indicated by the ``propto`` ("``prop``\ortional  ``to``")
+argument to function.
+
+If you are using an application such as an MCMC algorithm which requires
+gradients and only needs the log density up to a proportion, setting
+``propto=True`` will be at least as fast as setting ``propto=False``
+and is generally recommended (and the default value).
+
+However, in the case of the ``log_density`` function (which does not calculate
```
roualdes, line 20: I think it would be good to restate the "only needs the log density up to a proportion" bit again in this paragraph.

docs/internals/details.rst
```diff
@@ -0,0 +1,46 @@
...
+--------------------------------
+
+The log density function provided by a Stan model has
+the ability drop additive constants from the calculation.
+This is indicated by the ``propto`` ("``prop``\ortional  ``to``")
+argument to function.
+
+If you are using an application such as an MCMC algorithm which requires
+gradients and only needs the log density up to a proportion, setting
+``propto=True`` will be at least as fast as setting ``propto=False``
```
roualdes, line 17: Aren't calculations at most as fast with `propto = True`?  `propto = True` won't be faster, right?  Just maybe as fast or less fast.

docs/internals/details.rst
```diff
@@ -0,0 +1,46 @@
+Implementation Details
```
roualdes, line 1: In my opinion, the section title "Implementation Details" doesn't accurately describe what's being discussed here.  It's a Stan implementation detail, but not a BridgeStan implementation detail.  No doubt, the consequences of this Stan implementation detail affect BridgeStan.

roualdes (approved):
I like having things like this in BridgeStan cause it is overall details about Stan that affect our users and I definitely want to keep it here.  I just would rather offer the advice appropriate for our users and then point to Stan documentation.  But in the end, I don't know where this exists in Stan documentation.

docs/internals/details.rst
bob-carpenter: No.  If we are evaluating double values, then `propto=True` requires evaluating autodiff, whereas `propto=False` requires evaluating normalizing constants, but only with `double` inputs.  So it can go either way.

WardBrian:
I think this is essentially absent from the Stan documentation, since calculating gradients is essentially taken for granted everywhere in Stan

docs/internals/details.rst
WardBrian: Very happy to put it somewhere else or title it something different, I just did not know where to fit it in the current docs so I invented a new page

WardBrian:
Is there any place that would obviously lead the reader to the conclusion this discusses? e.g., that for the `log_density` function, `propto` can have dramatically different performance implications than it does for `log_density_gradient`?

bob-carpenter:
Good point---we talk about everything you would need to draw this conclusion yourself, but I don't think we ever connect the dots.  We probably should.  I added an issue for the *User's Guide* efficiency chapter:

https://github.com/stan-dev/docs/issues/692

docs/internals/details.rst
WardBrian: I went with "Stan implementation details"

PR #193, opened 2023-12-06
Python: allow ctypes pointers for in/out parameters

### LINKED ISSUE #190: Python callback into BridgeStan with C types

WardBrian:
Closes #190

This makes the minimal changes required to accept `ctypes.POINTER(ctypes.c_double)` types alongside numpy arrays for parameter arrays and out parameters

### DIFF
--- a/python/bridgestan/model.py
+++ b/python/bridgestan/model.py
@@ -14,24 +14,24 @@ from .compile import compile_model, windows_dll_path_setup
 from .util import validate_readable
 
 
-def wrapped_ndptr(*args, **kwargs):
+def array_ptr(*args, **kwargs):
     """
     A version of np.ctypeslib.ndpointer
-    which allows None (passed as NULL)
+    which allows raw ctypes pointers
     """
-    base = ndpointer(*args, **kwargs)
+    np_type = ndpointer(*args, **kwargs)
     ctypes_type = ctypes.POINTER(ctypes.c_double)
 
     def from_param(cls, obj):
         if isinstance(obj, ctypes_type):
             return ctypes_type.from_param(obj)
-        return base.from_param(obj)
+        return np_type.from_param(obj)
 
-    return type(base.__name__, (base,), {"from_param": classmethod(from_param)})
+    return type(np_type.__name__, (np_type,), {"from_param": classmethod(from_param)})
 
 
 FloatArray = npt.NDArray[np.float64]
-generic_double_array = wrapped_ndptr(dtype=ctypes.c_double, flags=("C_CONTIGUOUS"))
+generic_double_array = array_ptr(dtype=ctypes.c_double, flags=("C_CONTIGUOUS"))
 star_star_char = ctypes.POINTER(ctypes.c_char_p)
 c_print_callback = ctypes.CFUNCTYPE(None, ctypes.POINTER(ctypes.c_char), ctypes.c_int)
 
@@ -183,10 +183,10 @@ class StanModel:
         self._param_unc_num.argtypes = [ctypes.c_void_p]
 
         num_params = self._param_unc_num(self.model)
-        unc_p_double_array = wrapped_ndptr(
+        unc_p_double_array = array_ptr(
             dtype=ctypes.c_double, flags=("C_CONTIGUOUS"), shape=(num_params,)
         )
-        unc_p2_double_array = wrapped_ndptr(
+        unc_p2_double_array = array_ptr(
             dtype=ctypes.c_double,
             flags=("C_CONTIGUOUS"),
             shape=(num_params, num_params),
--- a/python/bridgestan/model.py
+++ b/python/bridgestan/model.py
@@ -23,14 +23,14 @@ def array_ptr(*args, **kwargs):
     ctypes_type = ctypes.POINTER(ctypes.c_double)
 
     def from_param(cls, obj):
-        if isinstance(obj, ctypes_type):
+        if isinstance(obj, (ctypes_type, ctypes.Array)):
             return ctypes_type.from_param(obj)
         return np_type.from_param(obj)
 
     return type(np_type.__name__, (np_type,), {"from_param": classmethod(from_param)})
 
 
-FloatArray = npt.NDArray[np.float64]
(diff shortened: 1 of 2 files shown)


python/bridgestan/model.py
```diff
@@ -13,8 +13,25 @@
 from .compile import compile_model, windows_dll_path_setup
 from .util import validate_readable
 
-FloatArray = npt.NDArray[np.float64]
-double_array = ndpointer(dtype=ctypes.c_double, flags=("C_CONTIGUOUS"))
+
+def array_ptr(*args, **kwargs):
+    """
+    A version of np.ctypeslib.ndpointer
```
bob-carpenter, line 19: I prefer the doc for a function to say what the function does.  This is going to return something so I'd phrase it as "Return < > given <...>" wherever possible.

Also use regular punctuation like periods at the end of sentences.

python/bridgestan/model.py
```diff
@@ -476,10 +503,7 @@ def param_unconstrain_json(
         dims = self.param_unc_num()
         if out is None:
             out = np.zeros(shape=dims)
-        elif out.size != dims:
```
bob-carpenter, line 479: Thanks for removing---this looks like it can't fail given the logic and that we've already checked it if it exists.

python/test/test_stanmodel.py
```diff
@@ -274,7 +275,7 @@ def test_param_unconstrain():
     c2 = bridge.param_unconstrain(b, out=scratch)
     np.testing.assert_allclose(a, c2)
     scratch_wrong = np.zeros(16)
-    with pytest.raises(ValueError):
+    with pytest.raises(ctypes.ArgumentError):
```
bob-carpenter, line 278: I also like the more specific exception types.

bob-carpenter (changes requested):
This looks good other than that I think the doc should be complete with argument declarations, return declaration, etc.

python/bridgestan/model.py
WardBrian: I rewrote it, let me know how it reads now

bob-carpenter:
Thanks for updating---looks good.

python/bridgestan/model.py
```diff
@@ -165,6 +186,18 @@ def __init__(
         self._param_unc_num.restype = ctypes.c_int
         self._param_unc_num.argtypes = [ctypes.c_void_p]
 
+        num_params = self._param_unc_num(self.model)
+        unc_p_double_array = array_ptr(
+            dtype=ctypes.c_double,
+            flags=("C_CONTIGUOUS", "WRITEABLE"),
```
roualdes, line 192: If I'm reading this right, it's great that you got out variables into Python code.

python/bridgestan/model.py
```diff
@@ -183,8 +216,8 @@ def __init__(
             ctypes.c_void_p,
             ctypes.c_int,
             ctypes.c_int,
-            double_array,
-            double_array,
+            generic_double_array,
+            generic_double_array,
```
roualdes, line 220: Maybe I missed the point on the type `unc_p_double_array`.  It seems like this should have the type associated with the variable name `unc_p_double_array` even though this name doesn't fit this case.  But shouldn't this out variable named `out` strictly be `WRITABLE`?

roualdes (approved):
This looks great.  Thanks for the elegant solution.

I'm stuck on the the variable named `unc_p_double_array` and the type stored by that variable.   It's not obvious to me which is more important to convey in our code, whether or not the output container is `WRITABLE` (the type is more important) or whether or not we're writing into a container that will hold variables on the unconstrained space (the variable name is more important).  But my being stuck shouldn't hold this up, we'd never get anywhere in this case :)

python/bridgestan/model.py
```diff
@@ -1,6 +1,6 @@
 import ctypes
 import warnings
-from os import PathLike, fspath
+from os import PathLike, fspath, write
```
roualdes, line 3: Where is `write` used?

roualdes (approved):
This is great. Thank you, Brian, for thinking about the ideas in my previous review, even if the words/sentences are hard to interpret.  I think your code here better represents our intentions, even if both codes have the same outcome.

PR #200, opened 2024-01-23
Use logos without text in docs

### LINKED ISSUE #199: logo icons
Same fix in two places:

1.  The icon used for web tabs shouldn't have "BridgeStan" under it.  It's way too small.

2.  On the doc pages, the icon used in upper left also shouldn't have "BridgeStan" under it because we include the name to the right.

The logo with the name under it will be good for standalone use.

WardBrian:
Closes #199


bob-carpenter (approved):
Thanks!

Issue #203, opened 2024-01-28
github.io page outdated


avehtari:
Page https://roualdes.github.io/bridgestan/latest/languages/r.html mentions `rng_seed`, but the code requires `seed`

bob-carpenter:
Closed with PR #204

PR #204, opened 2024-01-29
Update r.md from rng_seed to seed

bob-carpenter:
Changes `rng_seed` to `seed` in R documentation. 

Searched rest of doc to make sure it didn't appear elsewhere.

Fixes issue #203


WardBrian:
Thanks. This got missed way back in #98 

I'm going to look again into having R docs built automatically from the code like we have for everything else.

PR #205, opened 2024-01-29
Automate R docs build from roxygen comments

WardBrian:
Following on to #203 #204, this adds automated build support for the docs in R like we have in Julia and C++.

The online docs will now stay up to date with the documentation in the code. 

Like with Julia and C++, if the user doesn't have R locally the docs will still build, this section just will not update.


roualdes (approved):
This all seems good to me.  Thanks, Brian and Genentech apparently.

PR #211, opened 2024-02-13
improve panic error message on the Rust example

randommm:
a simple improvement suggestion: a better error message on the Rust example.

e.g.: in case of forgetting to compile the model with `STAN_THREADS=true`.


WardBrian (approved):
Thanks!

Issue #111, opened 2023-04-25
Deadlock on Windows when unloading libraries in several threads


aseyboldt:
While working on the rust backend we noticed problems that appear when stan modules are loaded and unloaded in separate threads.

On windows we see a deadlock under the following conditions:

Start three threads. Load stan module `A` in the first thread, while the other threads wait. Load module `B` in the second thread. Then, unload module `A` in the first thread and exit the thread. And last, try to load module `C` in the third thread. This will deadlock. From some debugging it seems that the second thread actually died somewhere while exiting.
This exact behavior seems to depend on the compiler or tbb version, because if we compile the stan model with clang or msvc instead of mingw and a recent TBB, we do not see the same issue (see https://github.com/roualdes/bridgestan/pull/105#issuecomment-1519113113)

This might be related to TBB, we found some issues that look at bit similar to this on the TBB bugtracker (https://github.com/roualdes/bridgestan/pull/105#issuecomment-1519121253).

A reproducer using the rust backend can be found here: https://github.com/aseyboldt/bridgestan/blob/main/rust/tests/create_models.rs#L236

There also seem to be some sporadic issues involving library unloading on linux as well however. With compilation settings `O=0` and enabled library unloading I observed a couple of sporadic segfaults when I ran the rust test. In a debugger it looked like the regular expression [here](https://github.com/stan-dev/stan/blob/dbc40e016c092209bd2262bfa02626ad4930fda6/src/stan/io/json/json_data_handler.hpp#L174) was not correctly initialized. I don't know for sure that this is related to unloading the libraries, but I've never seen that issue without it. And I think that since c++11 initialization of static variables should actually be [thread safe](https://en.cppreference.com/w/cpp/language/storage_duration#Static_local_variables).

(cc @WardBrian Feel free to add or modify the summary as you like)

aseyboldt:
I guess if I had checked the libloading issues a bit more thoroughly I might have saved myself quite a bit of time...

https://github.com/nagisa/rust_libloading/issues/41
https://github.com/nagisa/rust_libloading/issues/5

PR #212, opened 2024-02-15
implement BridgeStan download and module compilation on Rust

randommm:
Implement BridgeStan download and module compilation on Rust.

Fixes issue #100


rust/src/download_compile.rs
```diff
@@ -0,0 +1,108 @@
...
+
+        println!("Finished downloading BridgeStan");
+    }
+
+    Ok(bs_path_download_join_version)
+}
+
+/// Compile a Stan Model given a stan_file and the path to BridgeStan
+/// if None, then calls get_bridgestan_src() to download BridgeStan
+pub fn compile_model(stan_file: PathBuf, bs_path: Option<PathBuf>) -> Result<PathBuf> {
```
WardBrian, line 66: It would be nice to pass extra args like we do in the other language. In particular, adding `STANCFLAGS+=--include-paths=.` is needed for a lot of stan code "in the wild"

rust/tests/model.rs
WardBrian, line 1: Would you mind adding some tests specific to the new functionality?

WardBrian (changes requested):
Thank you for taking this on! A few thoughts on what is here so far

randommm:
> Thank you for taking this on! A few thoughts on what is here so far

okay, all done

WardBrian:
You may have to skip some of the compile tests on Windows for Rust, since we never unload libraries (#111) and Windows locks the file of loaded libraries

randommm:
i see, what about Mac? it seems to give some weird error that is hard to understand.

randommm:
there also seems to be some toolchain problem on Windows, lack of mingw, so it's unable to compile the example.

rust/README.md
```diff
@@ -9,44 +9,24 @@ natively from Rust.
 Internally, it relies on [`bindgen`](https://docs.rs/bindgen/) and
 [`libloading`](https://docs.rs/libloading/).
 
-## Compiling the model
-
-The Rust wrapper does not currently have any functionality to compile Stan models.
-Compiled shared libraries need to be built manually using `make` or with the Julia
-or Python bindings.
-
-For safety reasons all Stan models need to be installed with `STAN_THREADS=true`.
```
WardBrian, line 18: I think this and the later note about how an error will be thrown at runtime if it is not are still worth saying, if a user is using a library compiled from outside of their Rust program.

rust/.vscode/settings.json
WardBrian, line 1: Could you remove this file?

.github/workflows/main.yaml
```diff
@@ -290,5 +289,5 @@ jobs:
         run: |
           cargo clippy
           cargo fmt --check
-          cargo run --example=example
-          cargo test --verbose
+          cargo run --example=example --features compile-stan-model
+          cargo test --verbose --features compile-stan-model -- --test-threads=1
```
WardBrian, line 293: I'm a bit concerned that --test-threads=1 will have some impact on our other tests, but I don't recall entirely what it means. I think @aseyboldt and I discussed this when the rust bindings were initially developed

rust/src/download_compile.rs
```diff
@@ -0,0 +1,133 @@
...
+
+    let cmd = [
+        &[output.to_str().unwrap_or_default()],
+        make_args.as_slice(),
+        stanc_args.as_slice(),
+    ]
+    .concat();
+
+    println!("Compiling model");
+    let proc = std::process::Command::new("make")
```
WardBrian, line 117: I suspect this is why it was failing on Windows - the make command needed there is `mingw32-make`, not `make`

rust/src/bs_safe.rs
```diff
@@ -101,9 +103,17 @@ pub enum BridgeStanError {
     /// Setting a print-callback failed.
     #[error("Failed to set a print-callback: {0}")]
     SetCallbackFailed(String),
+    #[cfg(feature = "compile-stan-model")]
```
WardBrian, line 106: This is a question because I'm still relatively new to Rust: is putting specific parts of an enum behind a config a fine thing to do?

I guess the enum is already `non_exhaustive`, so it seems fine, but wanted to ask

rust/src/download_compile.rs
```diff
@@ -0,0 +1,133 @@
...
+
+    let bs_path_download_temp = homedir.join(".bridgestan_tmp_dir");
+    let bs_path_download = homedir.join(".bridgestan");
+
+    let bs_path_download_temp_join_version =
+        bs_path_download_temp.join(format!("bridgestan-{VERSION}"));
+    let bs_path_download_join_version = bs_path_download.join(format!("bridgestan-{VERSION}"));
+
+    if !bs_path_download_join_version.exists() {
+        println!("Downloading BridgeStan");
```
WardBrian, line 26: I know we do it in the other languages, but for some reason printing here feels less appropriate than in the scripting languages. OTOH, we do want to let people know what we're doing.

WardBrian (changes requested):
Thanks again. A few more comments below.

I'm also going to ask @aseyboldt if he would be willing to take a look at this

rust/README.md
randommm: Fair point, the reason i removed it is that if the model compile method is invoked, then it will recompile the model if it wasn't compiled with stan threads previously, but it's true that there is always a risk of someone not calling the compile method

.github/workflows/main.yaml
randommm: This is the number of threads that are ran in parallel, the reason I added this is because there is a race condition on the compile test because it deletes the .so file before proceding with the compilation and this can interfere with the other tests.

randommm: as an alternative i've added a "ignore" flag to this new test case (on the rust file), so then `cargo test` won't run it. So, then added a new line to main.yaml to run just this test after the other ones.

rust/src/download_compile.rs
randommm: I think you're right, I changed the code to use log::info instead of println

rust/src/bs_safe.rs
randommm: tricky question, according to my reading of this discussion https://www.reddit.com/r/rust/comments/xyik51/is_feature_gating_some_varaints_of_an_enum_a_bad/ it is acceptable if the enum is non_exhaustive, it's a controversial topic like, e.g. some people are against the use of box leak no matter what.

rust/Cargo.toml
```diff
@@ -11,10 +11,24 @@ homepage = "https://roualdes.github.io/bridgestan/latest/"
...
 thiserror = "1.0.40"
+ureq = { version = "2.7", optional = true }
+tar = { version = "0.4", optional = true }
+flate2 = { version = "1.0", optional = true }
+dirs = { version = "5.0", optional = true }
+path-absolutize = { version = "3.1", optional = true }
+log = { version = "0.4", optional = true }
+
+[features]
+compile-stan-model = ["ureq", "tar", "flate2", "dirs", "path-absolutize", "log"]
```
aseyboldt, line 22: I think we should not expose the optional dependencies as features, ie use the "dep:cratename" syntax mentioned here: https://doc.rust-lang.org/cargo/reference/features.html#optional-dependencies

rust/src/download_compile.rs
```diff
@@ -0,0 +1,139 @@
...
+    P: AsRef<Path>,
+{
+    let bs_path = match bs_path {
+        Some(path) => path.as_ref().to_owned(),
+        None => get_bridgestan_src()?,
+    };
+
+    let stan_file = stan_file
+        .as_ref()
+        .absolutize()
```
aseyboldt, line 88: Why not https://doc.rust-lang.org/std/fs/fn.canonicalize.html
Couldn't we drop the path_absolutize dep then?

rust/src/download_compile.rs
```diff
@@ -0,0 +1,139 @@
...
+        ));
+    }
+
+    // add _model suffix and change extension to .so
+    let output = stan_file.with_extension("");
+    let output = output.with_file_name(format!(
+        "{}_model",
+        output.file_name().unwrap_or_default().to_string_lossy()
+    ));
+    let output = output.with_extension("so");
```
aseyboldt, line 103: Maybe use https://doc.rust-lang.org/std/env/consts/constant.DLL_EXTENSION.html

rust/src/download_compile.rs
```diff
@@ -0,0 +1,139 @@
...
+
+    let stanc_args = [["--include-paths=."].as_slice(), stanc_args.as_slice()].concat();
```
aseyboldt, line 105: Why do we need the include-paths here?

rust/src/download_compile.rs
```diff
@@ -0,0 +1,139 @@
...
+    } else {
+        "make"
+    };
+    let proc = std::process::Command::new(make)
+        .args(cmd)
+        .current_dir(bs_path)
+        .env("STAN_THREADS", "true")
+        .output()
+        .map_err(|e| BridgeStanError::ModelCompilingFailed(e.to_string()))?;
+    info!("Finished compiling model");
```
aseyboldt, line 129: Maybe move to after the error check?

.github/workflows/main.yaml
```diff
@@ -290,5 +289,6 @@ jobs:
         run: |
           cargo clippy
           cargo fmt --check
-          cargo run --example=example
-          cargo test --verbose
+          cargo run --example=example --features compile-stan-model
+          cargo test --verbose --features compile-stan-model
+          cargo test --verbose --features compile-stan-model model_compiling -- --ignored
```
aseyboldt, line 294: I think we should test without the compile feature as well, just to make sure the `if cfg` are all correct.

aseyboldt (changes requested):
Sorry for the late reply...
@randommm Thank you for tackling this, I think I could use this in a couple of places myself.

I marked a couple relatively minor issues that I think should be pretty easy to fix.

I think I'd be fine with the PR if they are addressed, but there are two more other things that I'd also love to have, but they might be a bit more involved:

- Right now it is not obvious for a user of the API that some additional software is being downloaded. This has security implications, as that software could do anything it wants in theory. Maybe we could change the API a bit so that this is at least less hidden?
Maybe we also want to check the downloaded stan version somehow? If we tie it to a specific release we could just use a checksum I guess? Although I'm actually not sure how we'd best do that given that the file we want to download contains the repo, which would have to include the checksum...
- The model compilation function doesn't really work properly if used concurrently, even though nothing on the API side indicates that this is the case. I guess strictly speaking this probably isn't against the safety guarantees of rust, but it still seems to be against the spirit of those (I guess this is why you had to disable the test, or disable testing in parallel). Unfortunately this doesn't seem to be trivial to fix. We could just have a global Mutex that protects the compilation, but that would still fail if multiple independent processes try to compile models in parallel? (@WardBrian Is this something that has been looked into on the stan side?) Maybe we could use some locking mechanism to avoid this? So for instance we could add a `compiler_lock` file in the stan repository, and use fslock on that file to prevent other processes from using the stan installation at the same time. And then we could expose that on the rust side using something like:

```rust
bridgestan::Compiler::new()
    .ensure_download()?
    .lock()  // calls fslock on the lockfile inside the downloaded code
    .comile_model(path_to_model)?
// drop of the lock guard unlocks the lockfile
```

rust/src/download_compile.rs
randommm: this has some terrible "bug" on windows https://github.com/rust-lang/rust/issues/42869 It was hitting the CI/CD a few commits ago if you wish to check. I think std::path::absolute https://doc.rust-lang.org/std/path/fn.absolute.html would work here but is nightly.

rust/src/download_compile.rs
randommm: I thought about this too, but it seems that the Python interface always create a file .so extension, so i imagined this is intended?

rust/src/download_compile.rs
randommm: was included on Python binding so i thought it was usefull somehow https://github.com/roualdes/bridgestan/blob/b41f83b61571d4744ba5521a73d98a0c51fdd143/python/bridgestan/compile.py#L124

rust/src/bs_safe.rs
randommm: @aseyboldt could you share your opinion on this?

rust/Cargo.toml
randommm: well spotted, thanks! I've read this part of the Cargo book more than once but forgot about the implicit feature creation

WardBrian:
> We could just have a global Mutex that protects the compilation, but that would still fail if multiple independent processes try to compile models in parallel? (@WardBrian Is this something that has been looked into on the stan side?)

This is an unfortunate consequence of Stan really only supporting in-tree builds. This has been frustrating for [other reasons](https://github.com/stan-dev/cmdstan/issues/1175), I don't think this specific one has come up though. 

The right thing to do if concurrent builds are requested is to do them all in the same call to `make` and let `make`'s jobserver do it, but that seems very difficult to expose in a meaningful way. I guess the compile function could take a list of filenames, but that also seems cumbersome.

randommm:
> * I guess this is why you had to disable the test, or disable testing in parallel

Not exactly, it was just that i needed to delete the .so file before asking the function to compile the model, otherwise the function would skip compilation as the file already exists.

The problem then arises because other tests expect the .so file to exist in the meantime (they don't try to compile it first), so they will fail if they started after the file was deleted.

However, i think that if two processes tried to compile at the same time, i imagine that at least on Linux these wouldn't be a problem (on Windows, idk, conflict of two process opening the same file?).

randommm:
> > We could just have a global Mutex that protects the compilation, but that would still fail if multiple independent processes try to compile models in parallel? (@WardBrian Is this something that has been looked into on the stan side?)
> 
> This is an unfortunate consequence of Stan really only supporting in-tree builds. This has been frustrating for [other reasons](https://github.com/stan-dev/cmdstan/issues/1175), I don't think this specific one has come up though.
> 
> The right thing to do if concurrent builds are requested is to do them all in the same call to `make` and let `make`'s jobserver do it, but that seems very difficult to expose in a meaningful way. I guess the compile function could take a list of filenames, but that also seems cumbersome.

so right now, if two processes call make at the same time, bad things happen? I haven't tried but imagined that it wouldn't be bad on Linux

WardBrian:
I think it would be fine after the initial build? But if nothing has been successfully built yet I think you could definitely get a filesystem race on things like `main.o` that are built once and then re-used

rust/src/download_compile.rs
randommm: okay, see 86e9ee8e2997caa4b8cd713b2ecb54bc6ce0ca75

.github/workflows/main.yaml
randommm: done

randommm:
>   Maybe we also want to check the downloaded stan version somehow? If we tie it to a specific release we could just use a checksum I guess? Although I'm actually not sure how we'd best do that given that the file we want to download contains the repo, which would have to include the checksum...

Check commit fab21a7c2a6b0355 for a possible solution. The disadvantage is that will be download an outdated version of bridgestan, so not sure if this can cause a chicken and egg issue in the future in case ABI changes edge cases.

rust/src/bs_safe.rs
aseyboldt: I'm not sure either. I guess I'd rather not gate it, and have the variant in the enum all the time.

rust/src/download_compile.rs
aseyboldt: Oh no...
Maybe we could add a comment and mention that we can switch to https://doc.rust-lang.org/std/path/fn.absolute.html when that is stable?

rust/src/download_compile.rs
WardBrian: A lot of Stan programs in the wild want to be able to do includes relative to their current directory, which requires this extra flag. Most other interfaces to Stan pass it by default, so it's done here to match user expectations. 

Depending on how you invoke `make` you need to make sure the path you're including is the directory the model is in, not the bridgestan folder.

WardBrian:
> The disadvantage is that will be download an outdated version of bridgestan, so not sure if this can cause a chicken and egg issue in the future in case ABI changes edge cases.

I think this would definitely cause problems. Even lacking a major version bump, this would break even during minor version bumps if the new feature required new functions in the C++ layer.

You could instead do what we do for Julia:
We store the SHA of the [current version of the source,](https://github.com/roualdes/bridgestan/blob/main/julia/Artifacts.toml) before updating the Julia code itself in the release action
https://github.com/roualdes/bridgestan/blob/b41f83b61571d4744ba5521a73d98a0c51fdd143/.github/workflows/release.yaml#L61-L65

We could also start creating a tarball that is _only_ the C++ (this is probably a good idea for several reasons) which would make this chicken-egg problem easier

rust/src/download_compile.rs
aseyboldt: Makes sense. I think we need to change it then so that we use the original dir.

rust/src/download_compile.rs
randommm: okay, added the comments

rust/src/download_compile.rs
randommm: should be fixed now

randommm:
finally managed to fix the build on macos now!

randommm:
@WardBrian @aseyboldt made the API a little more clear now regarding that bridgestan is being download, and also easier to point to a custom bridgestan dir. I think that the checksum and the lock dir would be better fit as others issues/PRs.

rust/Cargo.toml
```diff
@@ -11,10 +11,24 @@ homepage = "https://roualdes.github.io/bridgestan/latest/"
...
 thiserror = "1.0.40"
+ureq = { version = "2.7", optional = true }
+tar = { version = "0.4", optional = true }
+flate2 = { version = "1.0", optional = true }
+dirs = { version = "5.0", optional = true }
+path-absolutize = { version = "3.1", optional = true }
+log = { version = "0.4", optional = true }
+
+[features]
+compile-stan-model = ["dep:ureq", "dep:tar", "dep:flate2", "dep:dirs", "dep:path-absolutize", "dep:log"]
```
WardBrian, line 22: Thanks for the changes. I'll let @aseyboldt comment on the ones he suggested, but this is looking pretty good from my end. 

I've given this some more thought and realized that almost all of these dependencies are not specific to _compilation_, but rather they're all for the "download the source if you don't have it" feature (right?).

Would it be reasonable to have _that_ be what is gated behind a feature? If the crate wasn't compiled with that and can't find the sources, it could unconditionally return the DownloadFailed error with a message saying so.

randommm: this is a good point, also helps on the security matter mentioned by @aseyboldt

rust/README.md
```diff
@@ -32,21 +29,25 @@ path = bridgestan.compile_model("stan_model.stan", make_args=["STAN_THREADS=true
...
+// The path to the Stan model
 let path = Path::new(env!["CARGO_MANIFEST_DIR"])
     .parent()
     .unwrap()
-    .join("test_models/simple/simple_model.so");
+    .join("test_models/simple/simple.stan");
+
+let bs_path = download_bridgestan_src().unwrap();
+// The path to the compiled model
+let path = compile_model(bs_path, path, vec![], vec![]).expect("Could not compile Stan model.");
```
WardBrian, line 49: Could we do something similar to the other examples and have this be a relative path (with the download option commented out as an example?)?

E.g.:
https://github.com/roualdes/bridgestan/blob/ef1f98694b9c6236e2d0b96eba025e864150e2de/julia/example.jl#L5-L9

rust/examples/example.rs
```diff
@@ -1,14 +1,25 @@
...
+    // The path to the Stan model
     let path = Path::new(env!["CARGO_MANIFEST_DIR"])
         .parent()
         .unwrap()
-        .join("test_models/simple/simple_model.so");
+        .join("test_models")
+        .join("simple")
+        .join("simple.stan");
+
+    let bs_path = download_bridgestan_src().unwrap();
```
WardBrian, line 19: Same as above: relative path pointing to the repo root?

rust/tests/model.rs
```diff
@@ -25,17 +25,83 @@ fn throw_data() {
...
+    let mut base = model_dir();
+    base.push(name);
+    let lib_path = base.join(format!("{}_model.so", name));
+    let stan_path = base.join(format!("{}.stan", name));
+    remove_file(lib_path).unwrap();
+
+    let homedir = dirs::home_dir().unwrap();
+    let bs_path_download = homedir.join(".bridgestan");
+    let bs_path_download_join_version =
+        bs_path_download.join(format!("bridgestan-{}", bridgestan::VERSION));
```
WardBrian, line 68: To more similarly match our other tests on other platforms, could we use the `$BRIDGESTAN` environment variable? This is particularly useful because it lets us test `VERSION+1`, where as this test assumes always that the compilation code will also work with the currently released version.

It also then does not need to rely on the download feature.

rust/tests/model.rs
```diff
@@ -25,17 +25,83 @@ fn throw_data() {
...
+    use common::model_dir;
+    use std::fs::remove_file;
+
+    let name = "stdnormal";
+    let mut base = model_dir();
+    base.push(name);
+    let lib_path = base.join(format!("{}_model.so", name));
+    let stan_path = base.join(format!("{}.stan", name));
+    remove_file(lib_path).unwrap();
+    let bs_path = download_bridgestan_src().unwrap();
```
WardBrian, line 98: Could this test be a bit more basic and just test that e.g. the path returned by downloading has a subfolder called `stan` and a file called `makefile`? I don't think there is anything necessary about testing compilation alongside the download, independent of the above test of just compilation?

rust/src/download.rs
```diff
@@ -0,0 +1,62 @@
...
+use std::{env::temp_dir, fs, path::PathBuf};
+use tar::Archive;
+
+/// Download and unzip the BridgeStan source distribution for this version
+/// to ~/.bridgestan/bridgestan-version
+pub fn download_bridgestan_src() -> Result<PathBuf> {
+    let homedir = dirs::home_dir().unwrap_or(temp_dir());
+
+    let bs_path_download_temp = homedir.join(".bridgestan_tmp_dir");
+    let bs_path_download = homedir.join(".bridgestan");
```
WardBrian, line 14: [optional, if not in this PR I will circle back around to do it eventually] It would be nice to have a similar warning to what we do in Python if a lot of these exist:

https://github.com/roualdes/bridgestan/blob/ef1f98694b9c6236e2d0b96eba025e864150e2de/python/bridgestan/compile.py#L73-L78

rust/src/compile.rs
```diff
@@ -0,0 +1,94 @@
...
+
+/// Compile a Stan Model given the path to BridgeStan and to a stan_file
+pub fn compile_model<P>(
+    bs_path: P,
+    stan_file: P,
+    stanc_args: Vec<&str>,
+    make_args: Vec<&str>,
+) -> Result<PathBuf>
+where
+    P: AsRef<Path>,
```
WardBrian, line 14: Question as much for @aseyboldt as @randommm: The `AsRef<Path>` generic is the canonical example I have seen for the [Non-Generic Inner Function](https://www.possiblerust.com/pattern/non-generic-inner-functions) pattern: do we care about doing something like that here? I have no sense of if this is useful outside things like stdlib functions you expected to get called in many different ways

WardBrian (commented):
The following are hopefully my final comments, focusing on some edge cases in the example and tests. 

I'd still like @aseyboldt do give another look over for any Rust nits I'm missing. But I think this is very good!

rust/src/compile.rs
randommm: fair point, idk, it was just a convenience to the user, but i think it's not worth the need for monomorphisations and such, so i've removed it.

rust/tests/model.rs
```diff
@@ -25,30 +25,81 @@ fn throw_data() {
...
+    let mut grad = [0f64];
+    let logp = model
+        .log_density_gradient(&theta[..], false, true, &mut grad[..])
+        .unwrap();
+    assert_ulps_eq!(logp, (2. * PI).sqrt().recip().ln() - 0.5);
+    assert_ulps_eq!(grad[0], -1f64);
+}
+
+#[cfg(feature = "download-bridgestan-src")]
+#[cfg(target_family = "unix")]
```
WardBrian, line 84: This should be platform independent

rust/tests/model.rs
```diff
@@ -25,30 +25,81 @@ fn throw_data() {
...
         .unwrap();
     assert_ulps_eq!(logp, (2. * PI).sqrt().recip().ln() - 0.5);
     assert_ulps_eq!(grad[0], -1f64);
 }
 
+// Note: this test does not require download-bridgestan-src feature
+// but it assumes that bridgestan_src is on the home folder
+// one way to have that assumption met is by running the example or
+// the model_downloading_and_compiling test before running this test
+#[cfg(target_family = "unix")]
```
WardBrian, line 50: Does this still require unix-only when it is run on its own due to the `[ignore]`?

WardBrian (approved):
This looks good to me. I left two tiny questions about the tests, and I would still appreciate if @aseyboldt could have a look before merging.

Thanks again!

PR #213, opened 2024-02-20
Move Julia TBB detection to instantiation time

WardBrian:
Unlike Python or R, Julia was setting up the Windows PATH at startup rather than at runtime. Among other things, this meant that calls to `set_bridgestan_path` couldn't influence this, since by the time that function was available, it was already too late.

This is in part thanks to a forum report: https://discourse.mc-stan.org/t/various-compilation-problems-with-various-interfaces/34155/5


roualdes (approved):
Looks good to me.  Thanks for looking into this Discourse report.

PR #218, opened 2024-03-15
Add flag `warn` to each constructor to control re-load warnings

### LINKED ISSUE #217: add no_warnings / quiet option to StanModel$new()
There are many use cases where there is a need to call `StanModel$new()` thousands of time with different data. After the second call there is always a warning message
```
Warning message:
In initialize(...) :
  Loading a shared object '/u/77/ave/unix/proj/cwolfe/multi_sim_missx_model.so' which is already loaded.
If the file has changed since the last time it was loaded, this load may not update the library!
```
Please add an option to turn this off. I could use R `suppressWarnings()`, but then I can miss other warnings. For example, if I have 
```
    suppressWarnings(mod2bs <- StanModel$new(lib = mod2bs_lib,
                                             data = to_stan_json(standati),
                                             seed = 1))
```
I may miss useful warnings from `to_stan_json()` function, and I would not like to create another variable to store the json before calling `StanModelnew()`

WardBrian:
Closes #217

### DIFF
--- a/R/R/bridgestan.R
+++ b/R/R/bridgestan.R
@@ -14,8 +14,9 @@ StanModel <- R6::R6Class("StanModel",
     #' @param seed Seed for the RNG used in constructing the model.
     #' @param stanc_args A list of arguments to pass to stanc3 if the model is not already compiled.
     #' @param make_args A list of additional arguments to pass to Make if the model is not already compiled.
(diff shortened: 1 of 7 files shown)


roualdes (approved):
Looks good.  Thanks much.

PR #227, opened 2024-04-15
Fix/226 julia manual download

### LINKED ISSUE #226: Consider moving Julia download away from LazyArtifacts
Hi -- just ran into this weird bug where an error is raised on Julia 1.10 while downloading bridgestan artifact inside a github CI Windows runner

https://github.com/Julia-Tempering/Pigeons.jl/actions/runs/8636249839/job/23678766648#step:8:1159

Interestingly this doesn't happen on Julia 1.9

https://github.com/Julia-Tempering/Pigeons.jl/actions/runs/8636249839/job/23678765917#step:8:1136

WardBrian:
This closes #226 and brings the Julia interface in line with the others by storing the bridgestan sources in `~/.bridgestan`

A couple things worth noting:

- Julia's `expanduser` function does nothing on Windows. I've added a helper function to get the home directory based on [what Python does](https://github.com/python/cpython/blob/7d0be7aea569b3bc9a3936501d7d32af87c70e73/Lib/ntpath.py#L363)
- This adds a few "dependencies": Tar, TOML, Downloads, and Inflate. I use the scare quotes because all except for Inflate are built in to Julia. Inflate, meanwhile, is [used in examples](https://github.com/python/cpython/blob/7d0be7aea569b3bc9a3936501d7d32af87c70e73/Lib/ntpath.py#L363) in Julia's documentation, and seems very stable.


WardBrian:
I confirmed that adding `--hard-dereference` to our packaging step produces a tarball which julia 1.6 can read. 

So, we can either:

- merge this while it is failing, knowing the next release will actually be fine
- merge just the `release.yml` change to add that flag, make a small release, and then merge the Julia changes

roualdes (approved):
This looks good.  And thanks for finding a fix that includes Julia 1.6.  Since 1.6 is the LTS version, it seems right to include it.

I guess it's better to get this fix out there, and then tidy 1.6 issues later, since this PR started based on a reported issue with Julia 1.10, and not an issue related to 1.6.  Maybe that's wrong, but that's my thought.

PR #228, opened 2024-04-22
Update Julia doc to cross link, remove mention of Artifact

WardBrian:
Removes a extraneous mention of `Artifacts.jl` after #227, and uses the `@ref` syntax to make Julia functions have clickable links to their "see also" items.


roualdes (approved):
Sorry I missed this.  Looks good to me.  Nice to get the links in the Julia doc.  So now just R is missing this, ya?

PR #229, opened 2024-04-26
Clean up our usage of Bool in FFI

WardBrian:
I recently ran into [some issues](https://github.com/JuliaLang/julia/issues/54257) which originated from improperly passing `bool` arguments as `Int32`s. 

This appears to be fine on most architectures, especially in our case where they're small functions passing arguments on the stack, but it still seemed like something worth fixing up.

I also was made aware of the much more compact `@ccall` Julia macro that makes it much clearer what types are associated with which arguments

### DIFF
--- a/.github/workflows/main.yaml
+++ b/.github/workflows/main.yaml
@@ -8,7 +8,7 @@ on:
   workflow_dispatch: {}
 
 env:
-  CACHE_VERSION: 0
+  CACHE_VERSION: 1
 
 # only run one copy per PR
 concurrency:
@@ -270,12 +270,6 @@ jobs:
           path: ./test_models/
           key: ${{ hashFiles('**/*.stan', 'src/*', 'stan/src/stan/version.hpp', 'Makefile') }}-${{ matrix.os }}-v${{ env.CACHE_VERSION }}
 
-      - name: Install LLVM and Clang
-        uses: KyleMayes/install-llvm-action@v2
-        with:
-          version: "15.0"
-          directory: ${{ runner.temp }}/llvm
-
       - name: Set up TBB
         if: matrix.os == 'windows-latest'
         run: |
@@ -284,9 +278,6 @@ jobs:
       - name: Run rust tests
         working-directory: ./rust
         timeout-minutes: 60
-        env:
-          LIBCLANG_PATH: ${{ runner.temp }}/llvm/lib
-          LLVM_CONFIG_PATH: ${{ runner.temp }}/llvm/bin/llvm-config
         run: |
           cargo clippy
           cargo fmt --check
--- a/julia/src/model.jl
+++ b/julia/src/model.jl
@@ -1,3 +1,4 @@
+using Base.Libc.Libdl: dlsym, dllist, dlopen
 
 struct StanModelStruct end
 
@@ -54,7 +55,7 @@ mutable struct StanModel
             lib = compile_model(lib; stanc_args, make_args)
         end
 
-        if warn && in(abspath(lib), Libc.Libdl.dllist())
+        if warn && in(abspath(lib), dllist())
             @warn "Loading a shared object '" *
                   lib *
                   "' which is already loaded.\n" *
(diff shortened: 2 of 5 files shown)


src/bridgestanR.cpp
```diff
@@ -23,23 +23,24 @@ void bs_model_info_R(bs_model** model, char const** info_out) {
 }
 void bs_param_names_R(bs_model** model, int* include_tp, int* include_gq,
                       char const** names_out) {
-  *names_out = bs_param_names(*model, *include_tp, *include_gq);
+  *names_out = bs_param_names(*model, (*include_tp != 0), (*include_gq != 0));
```
roualdes, line 26: Why not change the function parameter from `int* include_tp` to `bool* include_tp`?

`bridgestan/R/R/bridgestan.R` (that's funny) is passing arguments as `as.logical(include_tp), as.logical(include_gq)` so aren't we still oddly juggling ints and bools?

.github/workflows/main.yaml
```diff
@@ -270,12 +270,6 @@ jobs:
           path: ./test_models/
           key: ${{ hashFiles('**/*.stan', 'src/*', 'stan/src/stan/version.hpp', 'Makefile') }}-${{ matrix.os }}-v${{ env.CACHE_VERSION }}
 
-      - name: Install LLVM and Clang
```
roualdes, line 273: Nice.  I love being able to remove things and have everything still work.

julia/src/model.jl
```diff
@@ -72,31 +73,24 @@ mutable struct StanModel
...
-
-        stanmodel = ccall(
-            Libc.Libdl.dlsym(lib, "bs_model_construct"),
-            Ptr{StanModelStruct},
-            (Cstring, UInt32, Ref{Cstring}),
-            data,
-            seed,
-            err,
-        )
+        stanmodel = @ccall $(dlsym(lib, :bs_model_construct))(
```
roualdes, line 79: What does `@ccall ...` do to error messages, when wrongly typed arguments are passed? I mean, this question should not stop this change from going in, but I've had bad experiences with error messages coming from within a function call which lives inside a macro.

roualdes (commented):
Thanks for catching this and following it through.  Python looks good.  One interest based question in Julia, and one real question from R. I just don't understand why the change happened as it did in R, but I'm probably just overlooking something.

src/bridgestanR.cpp
WardBrian: I checked this again this morning, and R's documentation states that `int*` is correct:

```
     The mapping of the types of R arguments to C or Fortran arguments
     is

       R          C                Fortran          
       integer    int *            integer          
       numeric    double *         double precision 
       - or -     float *          real             
       complex    Rcomplex *       double complex   
       logical    int *            integer   
```

The reason is that logical values in R can not just be `TRUE` and `FALSE`, but also `NA`:

```
     Logical values are sent as ‘0’ (‘FALSE’), ‘1’ (‘TRUE’) or ‘INT_MIN
     = -2147483648’ (‘NA’, but only if ‘NAOK = TRUE’), and the compiled
     code should return one of these three values: however non-zero
     values other than ‘INT_MIN’ are mapped to ‘TRUE’.
```

julia/src/model.jl
WardBrian: Are you asking if an end user messes up the types or if we do? I think our functions would stop them before they hit this, but maybe not if we ourselves made a mistake?

roualdes: Ya, sorry, that wasn't clear.  I had in mind something like a user passing in a wrongly typed variable into `param_constrain()` or `log_density()` or something, but I see now / remember that all the functions I had in mind have typed parameters themselves.  So any type errors will catch there, and not here.

src/bridgestanR.cpp
roualdes: I did not know this.  Thanks.

roualdes (approved):
Great.  Thanks much for tidying up our use of bools across languages (darn you R) and platforms.

PR #232, opened 2024-05-03
R: Warn if error is too long to print properly

WardBrian:
R has a [maximum length for warnings/error messages,](https://stackoverflow.com/questions/50387667/how-to-prevent-truncation-of-error-messages-in-r) controlled by `options(warning.length)`. This has lead to some confusion before, see [1](https://discourse.mc-stan.org/t/post-hoc-prediction-use-samples-from-fitted-model-in-a-prediction-task/34319/29?u=wardbrian), [2](https://discourse.mc-stan.org/t/various-compilation-problems-with-various-interfaces/34155/7?u=wardbrian)

This adds a check to places where we know there could be long error messages generated by BridgeStan and adds a warning if we know they will be truncated. This warning is reported alongside the error and should help people debug

### DIFF
--- a/R/R/bridgestan.R
+++ b/R/R/bridgestan.R
@@ -347,6 +347,9 @@ handle_error <- function(lib_name, err_msg, err_ptr, function_name) {
     return(paste("Unknown error in", function_name))
   } else {
     .C("bs_free_error_msg_R", as.raw(err_ptr), PACKAGE = lib_name)
+    if (getOption("warning.length") < nchar(err_msg)) {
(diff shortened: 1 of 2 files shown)


roualdes (approved):
Good idea.  Do we have a recommended first guess on what `warning.length` they could try?

Maybe we could say `options(warning.length = 1000)` or whatever our first best guess is, if not 1000.

WardBrian:
1000 is the default, the linked SO question suggests values above 8000 don’t do anything. 

We could also just print it and have the error message say “hey, we printed it just now”, but I think that would interact poorly with anyone trying to intercept errors

PR #233, opened 2024-05-03
Support for USER_HEADER in build system

WardBrian:
This ports the portion of the CmdStan makefile that allows for easier inclusion of external C++ functions:
https://mc-stan.org/docs/cmdstan-guide/external_code.html

Essentially, if the `--allow-undefined` option is passed to stanc3, we then look for a header file to be included during compilation. By default this is `user_header.hpp` in the same folder as the Stan model, but it can be changed by setting `USER_HEADER`.

I'm not sure where is the best place to slot in documentation for this.

### DIFF
--- a/docs/getting-started.rst
+++ b/docs/getting-started.rst
@@ -165,6 +165,19 @@ to set :makevar:`CPPFLAGS` in :file:`make/local`:
 
     CPPFLAGS+=-DSTAN_MATH_CONSTRAINT_TOLERANCE=1e-5
 
+
+Using External C++ Code
+_______________________
+
+BridgeStan supports the same `capability to plug in external C++ code as CmdStan <https://mc-stan.org/docs/cmdstan-guide/external_code.html>`_.
+
+Namely, you can declare a function in your Stan model and then define it in a separate C++ file.
+This requires passing the ``--allow-undefined`` flag to the Stan compiler when building your model.
+The :makevar:`USER_HEADER` variable must point to the C++ file containing the function definition.
+By default, this will be the file :file:`user_header.hpp` in the same directory as the Stan model.
+
+For a more complete example, consult the `CmdStan documentation <https://mc-stan.org/docs/cmdstan-guide/external_code.html>`_.
+
 Using Older Stan Versions
 __________________________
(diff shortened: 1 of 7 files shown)


roualdes:
Great addition.  Thanks.

I suppose the Customizing BridgeStan part of the doc?  With a link to the CmdStan User's Guide.

Any idea why the Rust tests failed?

WardBrian:
Rust tests try to load all the modules from the folder, so I think we just need to exclude this the same way we exclude syntax_error. 

Customizing BridgeStan is probably the best section, though it's more per-model that most of the other recommendations there

roualdes:
Let me know when you are ready for me to look this over again.  Thanks.

WardBrian:
@roualdes I think it's ready for another look!

docs/getting-started.rst
```diff
@@ -165,6 +165,19 @@ to set :makevar:`CPPFLAGS` in :file:`make/local`:
...
 
+
+Using External C++ Code
+_______________________
+
+BridgeStan supports the same `capability to plug in external C++ code as CmdStan <https://mc-stan.org/docs/cmdstan-guide/external_code.html>`_.
+
+Namely, you can declare a function in your Stan model and then define it in a separate C++ file.
+This requires passing the ``--allow-undefined`` flag to the Stan compiler when building your model.
+The :makevar:`USER_HEADER` variable must point to the C++ file containing the function definition.
```
roualdes, line 176: Should we say explicitly that it should be a C++ header file?  We only reference it being a `.hpp` file in the makevar variable name and the example file name.

roualdes (approved):
Just a quick question, but overall looks good to me.  A nice quality of life feature to add to BridgeStan.  Thanks.

docs/getting-started.rst
WardBrian: It technically does not need to be a header (or even a `.hpp`, in theory a `.txt` would also work, since the compiler is just pasting it in), as long as it's code that can be included into the model's translation unit it would be fine. 

Personally I think the name `USER_HEADER` is evocative enough

PR #236, opened 2024-06-17
Add a download flag to get_bridgestan_path

### LINKED ISSUE #220: Ability to check if bridgestan sources are already available
- Why `get_bridgestan_path()` is not exported?
- The behavior is strange. I have not set `BRIDGESTAN` environment variable, and get message
```
BridgeStan not found at location specified by $BRIDGESTAN environment variable, downloading version 2.4.0 to  /u/77/ave/unix/.bridgestan/bridgestan-2.4.0"
```
but `/u/77/ave/unix/.bridgestan/` directory does already have working bridgestan installation, so it was able to find the correct directory, but not able to check if there is already an installation, and it goes and downloads something without giving a change to set the path if I would use some other path than the default.

WardBrian:
Closes #220 (@avehtari)

This adds a flag `download` to all the interfaces' `get_bridgestan_path` functions. The default is true, which is the previous behavior. Setting to `false` allows you to inspect the current value without triggering a download if it is not found. 

Internally we always use the default value, but some users may want to check if their custom installation is working without triggering a download if it isn't.

### DIFF
--- a/julia/src/compile.jl
+++ b/julia/src/compile.jl
@@ -4,7 +4,7 @@ function get_make()
 end
 
 
-function validate_stan_dir(path::AbstractString)
+function verify_bridgestan_path(path::AbstractString)
     if !isdir(path)
         error("Path does not exist!\n$path")
     end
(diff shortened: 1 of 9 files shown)


julia/src/compile.jl
```diff
@@ -17,23 +17,31 @@ end
...
 
 Return the path the the BridgeStan directory.
 
-If the environment variable `BRIDGESTAN` is set, this will be returned.
-Otherwise, this function downloads a matching version of BridgeStan under
-a folder called `.bridgestan` in the user's home directory.
+If the environment variable `\$BRIDGESTAN` is set, this will be returned.
+
+If `\$BRIDGESTAN` is not set and `download` is true, this function downloads
+a matching version of BridgeStan under a folder called `.bridgestan` in the
```
roualdes, line 27: I've forgotten about what matching means here, so maybe we could make it more explicit?

roualdes (approved):
This looks good.  Thanks for addressing this request.  Happy to move forward as is.

Here's some nitpicks, but they're more like things I noticed while looking over this PR.  

* Is there anything I'm missing about having lines like`path = CURRENT_BRIDGESTAN` inside the try-catch versus outside?
* I see now that R/Python have `verify_bridgestan_path`, while Julia has `validate_bridgestan_path`, but it seems like they do the same thing.

roualdes (approved):
Thanks much!

PR #241, opened 2024-07-26
Input size checks

### LINKED ISSUE #239: parameter sizes not checked in log_density_gradient
The `log_density_gradient` function doesn't check we get the right number of parameters in a call. 

Here's a simple reproducible example.

`bs.stan`
```stan
parameters {
  vector<lower=0, upper=1>[2] alpha;
}
```

Running it with too small, just right, and too large array of parameters:

```python
import bridgestan as bs
import numpy as np
model = bs.StanModel('bs.stan')
lp1, grad1 = model.log_density_gradient(np.array([0.3]));
lp2, grad2 = model.log_density_gradient(np.array([0.3, 0.5]));
lp3, grad3 = model.log_density_gradient(np.array([0.3, 0.5, 0.1]))
print(f"{lp1=} {lp2=} {lp3=}")
print(f"{grad1=} {grad2=} {grad3=}")
```

No errors, just this output, which looks like it ignores extra parameters and treats missing parameters as having value 0.

```python
>>> print(f"{lp1=} {lp2=} {lp3=}")
lp1=-2.795004850056945 lp2=-2.8568644572972675 lp3=-2.8568644572972675

>>> print(f"{grad1=} {grad2=} {grad3=}")
grad1=array([-0.14888503,  0.        ]) grad2=array([-0.14888503, -0.24491866]) grad3=array([-0.14888503, -0.24491866])
```

I think it'd be useful to check input size here.  This came to my attention as a bug in someone's code they were showing me that BridgeStan didn't catch.

WardBrian:
Would close #239 

I only timed this in Python, where I was not really able to consistently time a difference even in a very simple log density. 

Note: Some existing code will fail after this is merged, even if that code is technically buggy. Worth noting in the release notes.

### DIFF
--- a/R/R/bridgestan.R
+++ b/R/R/bridgestan.R
@@ -59,6 +59,9 @@ StanModel <- R6::R6Class("StanModel",
       }
       private$model <- ret$ptr_out
 
+      # pre-compute to avoid repeated work in bounds checks
+      private$unc_dims <- self$param_unc_num()
+
       model_version <- self$model_version()
       if (packageVersion("bridgestan") != paste(model_version$major, model_version$minor, model_version$patch, sep = ".")) {
         warning(paste0("The version of the compiled model does not match the version of the R library. ",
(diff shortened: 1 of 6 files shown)


roualdes (approved):
Sorry for the super slow response on this.  My bad.

This looks good to me.  

It's nice in Python that we're getting some good use out of the `array_ptr` class.  Keeps the implementation cleaner than in Julia or R.

Does the Rust implementation already do size checks?

WardBrian:
Rust already did, yes

PR #248, opened 2024-10-16
Add example of loading at runtime from C

WardBrian:
Recently a few requests have come in around using BridgeStan from C/C++, but in a way that is more closely aligned with how it is used in the other languages, e.g. providing a path to the library, rather than re-linking your code.

This adds an alternative example that does just that. It also adds some extra config so we can run some of these examples in the Windows CI -- the statically linked one isn't working for me, so I left it out for now still.

### DIFF
--- a/c-example/Makefile
+++ b/c-example/Makefile
@@ -4,11 +4,17 @@ include ../Makefile
 
 MODEL?=full
 CC = gcc
+# .dll on Windows, .so on Linux
+ifeq ($(OS_TAG),windows)
+	DLL = dll
+else
+	DLL = so
+endif
 
-../test_models/$(MODEL)/lib$(MODEL)_model.so: ../test_models/$(MODEL)/$(MODEL)_model.so
-	cp ../test_models/$(MODEL)/$(MODEL)_model.so ../test_models/$(MODEL)/lib$(MODEL)_model.so
+../test_models/$(MODEL)/lib$(MODEL)_model.$(DLL): ../test_models/$(MODEL)/$(MODEL)_model.so
+	cp ../test_models/$(MODEL)/$(MODEL)_model.so ../test_models/$(MODEL)/lib$(MODEL)_model.$(DLL)
 
-example$(EXE): example.c ../test_models/$(MODEL)/lib$(MODEL)_model.so
+example$(EXE): example.c ../test_models/$(MODEL)/lib$(MODEL)_model.$(DLL)
 	$(CC) -c -I ../src example.c -o example.o
 	$(LINK.c) -o example$(EXE) example.o -Wl,-rpath ../test_models/$(MODEL) -L ../test_models/$(MODEL) -l$(MODEL)_model
 	$(RM) example.o
@@ -30,4 +36,4 @@ example_static$(EXE):  example.c ../test_models/$(MODEL)/$(MODEL)_model.a
 	$(RM) example.o
 
 example_runtime$(EXE): runtime_loading.c
-	$(CC) -I ../src runtime_loading.c -o example_runtime$(EXE) -ldl
+	$(CC) -I ../src runtime_loading.c -o example_runtime$(EXE)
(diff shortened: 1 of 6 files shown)


c-example/Makefile
```diff
@@ -28,3 +34,6 @@ example_static$(EXE):  example.c ../test_models/$(MODEL)/$(MODEL)_model.a
 	$(CC) -c -I ../src example.c -o example.o
 	$(LINK.cpp) -o example_static$(EXE) example.o ../test_models/$(MODEL)/$(MODEL)_model.a $(LDLIBS) $(LIBSUNDIALS) $(MPI_TARGETS) $(TBB_TARGETS)
 	$(RM) example.o
+
+example_runtime$(EXE): runtime_loading.c
```
roualdes, line 38: Is it reasonable to add `../test_models/$(MODEL)/lib$(MODEL)_model.$(DLL)` here?

c-example/README.md
```diff
@@ -7,7 +7,9 @@ This shows how one could write a C program which calls BridgeStan.
 Any compiled language with a C foreign function interface and
 the ability to link against C libraries should be able to work similarly.
 
-## Usage with dynamic linking
+## Binding a specific model at build type
```
roualdes, line 10: type or time?

c-example/runtime_loading.c
```diff
@@ -0,0 +1,90 @@
...
+
+#ifdef _WIN32
+// hacky way to get dlopen and friends on Windows
+
+#include <libloaderapi.h>
+#include <errhandlingapi.h>
+#define dlopen(lib, flags) LoadLibraryA(lib)
+#define dlsym(handle, sym) (void*)GetProcAddress(handle, sym)
+
+char* dlerror() {
```
roualdes, line 12: Honestly, I don't know how important this is, and my guess is not that important, but it came to mind so I'm saying something.  Should the `dlerror` signature be `void dlerror(char* err)`, so as to free memory outside of the function?

roualdes (approved):
Great idea, thanks, Brian.  Just a couple of questions.

c-example/Makefile
WardBrian: It would cause `example_runtime` to be re-built more often, which feels like it goes against the pedagogical purpose of the example. Strictly speaking, there is no reason that any bridgestan library needs to have been compiled before this, which is what that dependency would be saying

c-example/README.md
WardBrian: time, my bad

c-example/runtime_loading.c
WardBrian: This only exists to mirror the `dlerror` that is built-in on Linux/Mac, so we're constrained to the same signature. It definitely does leak memory, but since all we do in this tiny example is call it right before quitting the program, it doesn't matter for us really

c-example/Makefile
roualdes: > goes against the pedagogical purpose of the example

Right, good point. 

Is it reasonable to leave the code suggestion in the Readme as is, then add a sentence or two explaining that `example_runtime` the executable can be compiled, but not run without a `.so`?

c-example/runtime_loading.c
roualdes: Yep, that makes and sense and is fair.  I was just (probably overly) worried about our example being taken too literally.

WardBrian: Yeah, hopefully someone doing this "for real" on Windows would try a bit harder... my comment in the code calling this hacky was mostly out of this same concern

c-example/Makefile
WardBrian: I updated the readme block:

>
```shell
make example_runtime
# unlike above, this did not automatically build a model, since it
# was not needed to _build_, but we still need one to _run_ the program
make ../test_models/full/full_model.so
./example_runtime ../test_models/full/full_model.so
```

roualdes (approved):
Thanks!

PR #257, opened 2024-12-13
Bump min Julia version to new LTS (1.10)

WardBrian:
https://julialang.org/downloads/#long_term_support_release

This wouldn't take effect until the next release, 2.6.0 was just released with the existing 1.6 version bound.


roualdes (approved):
Ya, this seems like a reasonable thing to do. Looks good to me.  Thanks.

PR #259, opened 2024-12-23
Fix rust send sync

### LINKED ISSUE #258: Rust threadsafety annotation changed in release 2.6.0
In version 2.5.0 the StanLibrary struct was marked as `Send` and `Sync`:
https://docs.rs/bridgestan/2.5.0/bridgestan/struct.StanLibrary.html#synthetic-implementations
But in 2.6.0 the same struct is marked as not `Send` and `Sync`:
https://docs.rs/bridgestan/latest/bridgestan/struct.StanLibrary.html#synthetic-implementations

If some entry on the c-side changed from const to non-const, that might explain why this is suddenly different, but I couldn't find any such change.
So maybe the update of bindgen triggered this?

I *think* `StanLibrary` is in fact threadsafe, I can't see any reason why different threads couldn't create models at the same time? But maybe I'm missing something on the C++ side...

If we think it should be safe, I can make a simple PR that informs rust about this.

aseyboldt:
Fixes #258 by explicitly marking the StanLibrary as Send + Sync.

This also adds an entry to the github workflow, which should automatically find most backward incompatible changes that break semver.

### DIFF
--- a/.github/workflows/main.yaml
+++ b/.github/workflows/main.yaml
@@ -300,3 +300,7 @@ jobs:
           # run all tests with feature download-bridgestan-src
           cargo test --verbose --all-features
           cargo test --verbose model_compiling -- --ignored
+      - name: Check semver for rust interface
+        uses: obi1kenobi/cargo-semver-checks-action@v2
+        with:
+          manifest-path: ./rust/Cargo.toml
--- a/rust/src/bs_safe.rs
+++ b/rust/src/bs_safe.rs
@@ -25,6 +25,10 @@ pub struct StanLibrary {
     id: u64,
 }
 
+// Stan libraries are thread safe.
+unsafe impl Send for StanLibrary {}
+unsafe impl Sync for StanLibrary {}
+
 // To work around a bug where unloading a library
 // can lead to deadlocks.
 //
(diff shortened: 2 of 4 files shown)


aseyboldt:
I think I found the reason (or part of it), why rust thought this wasn't threadsafe.
The library contains version numbers as `int`, not `const int`, so two threads could modify the version number concurrently in an unsafe way.
This only showed up now, because the version numbers were not exposed to rust in the old brindgen version, but with 0.71 they are.
I changed the versions to const, and I think this *should* now automatically make Send + Sync threadsafe, but there seems to be a bindgen issue (https://github.com/rust-lang/rust-bindgen/issues/3059), and with this it is still exported as a `mut int`.
So I left the explicit Send + Sync implementation in. Since the wrapper never modifies the version (that would be fun) I think this is fine.

WardBrian:
Good catch, and I think this is the correct fix. 

We can release a 2.6.1 to patch this. The merge of #

WardBrian:
(Sorry, fat thumbs!)

Good catch, and I think this is the correct fix. 

We can release a 2.6.1 to patch this. The merge of #257 makes this a bit annoying, since we probably don’t want to change the Julia LTS in a minor version. We can try to rig up something to skip releasing Julia for this patch or I can do some cherry picking on another branch after the holidays

aseyboldt:
Sounds good, thanks.
And (slightly early) Merry Christmas :-)

PR #264, opened 2025-01-21
Start using sphinxcontrib-rust for Rust API documentation

WardBrian:
There is a relatively newer projected called [sphinxcontrib-rust](https://gitlab.com/munir0b0t/sphinxcontrib-rust) that allows us to do something better than just link to docs.rs

Preview: https://brianward.dev/bridgestan/latest/languages/rust.html#api-reference

I also fixed a few other links in the docs while I was going.


roualdes:
This looks great to me.  I'd happily approve, but I've hardly had a hand in the Rust interface, so it's not obvious that I should be the final approval here.

The drop down example should say example.rs instead of example.R.

WardBrian:
> so it's not obvious that I should be the final approval here.

I don't know how much doc changes like this need review besides things like catching typos (which you've already done!)

docs/languages/c-api.rst
```diff
@@ -16,8 +16,7 @@ Example Program
 Two example programs are provided alongside the BridgeStan source in :file:`c-example/`.
 Details for building the example can be found in :file:`c-example/Makefile`.
 
-The first assumes you wish to link a specific model into the program,
-and the second demonstrates how to load a model at runtime
+The if you wish to link a specific model into the program:
```
roualdes, line 19: typo: the if

roualdes (approved):
One small typo, otherwise looks good.  Thanks.

PR #274, opened 2025-04-28
Update related projects section of README

WardBrian:
A few new papers and software packages I came across


roualdes (approved):
Nice.  Thanks for keeping this updated.

PR #275, opened 2025-05-05
Forward declare model and rng class when including bridgestan.h in C++

WardBrian:
This prevents you from needing to have all of Stan on your include path to be able to use bridgestan.h in C++

### DIFF
--- a/src/model_rng.cpp
+++ b/src/model_rng.cpp
@@ -11,13 +11,10 @@
 #ifdef BRIDGESTAN_AD_HESSIAN
 #include <stan/math/mix.hpp>
 #endif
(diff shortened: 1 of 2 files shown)


roualdes (approved):
Ya, this looks OK to me.  You've tested this against the code you're writing up for Bob?

There are some changes to `rust/tests/model.rs`, but I assume we're just throwing these in arbitrarily cause they should happen and need to go somewhere, right?  They otherwise seem unrelated.

WardBrian:
> They otherwise seem unrelated.

Yeah, they were just silencing some deprecation warnings that my IDE was bothering me about...

PR #276, opened 2025-05-06
Clean up, split model_rng source, avoid unnecessary .cpp file

WardBrian:
This is something that has bothered me for a bit but I didn't see a better fix for it until after #275.

Basically, we (currently) have a model_rng.hpp, and a model_rng.cpp. There are a few reasons to do this, but none of them seem to apply to this code:

- Implementation hiding (from who?)
- Reducing the number of things the compiler needs to process (the .hpp was already including something from `<stan/`, so it needed everything anyway)
- Separate compilation (which we aren't doing, and wouldn't really benefit from for this)

Because we aren't doing separate compilation, we also had some nonstandard stuff like `#include`ing a .cpp file in another .cpp file. I'm pretty sure we had avoided it so far, but this is a pretty good way to end up with One Definition Rule violations if you're not careful. 

So, this PR just stops doing that split. The entire definition is available in the header. None of the code is different, just in different files. The only code that did change was in bridgestan.cpp, the handling of the outstream global was updated to avoid needing to define a variable in the header (and, in the process, fix a potential race condition)


roualdes (approved):
Looks good to me and the tests pass.

PR #277, opened 2025-05-08
Maybe: clean up error handling

WardBrian:
This removes a lot of boilerplate for our exception handling strategy by using a helper function that takes in a lambda. 

By forcing the function to inline, this actually generates more or less identical assembly: https://godbolt.org/z/3sqWoscnn

In this is too fancy, I understand 😃

### DIFF
--- a/src/bridgestan.cpp
+++ b/src/bridgestan.cpp
@@ -3,10 +3,12 @@
 #include "rng.hpp"
 #include "callback_stream.hpp"
 #include "version.hpp"
+#include "util.hpp"
 
 #include "bridgestanR.cpp"
 
-#include <sstream>
+
+using bridgestan::handle_errors;
 
 const int bs_major_version = BRIDGESTAN_MAJOR;
 const int bs_minor_version = BRIDGESTAN_MINOR;
@@ -14,27 +16,8 @@ const int bs_patch_version = BRIDGESTAN_PATCH;
 
 bs_model* bs_model_construct(const char* data, unsigned int seed,
                              char** error_msg) {
-  try {
-    return new bs_model(data, seed);
-  } catch (const std::exception& e) {
-    if (error_msg) {
-      std::stringstream error;
-      error << "construct(" << (data == nullptr ? "NULL" : data) << ", " << seed
-            << ")"
-            << " failed with exception: " << e.what() << std::endl;
-      *error_msg = strdup(error.str().c_str());
-    }
-  } catch (...) {
-    if (error_msg) {
-      std::stringstream error;
-
-      error << "construct(" << (data == nullptr ? "NULL" : data) << ", " << seed
-            << ")"
-            << " failed with unknown exception" << std::endl;
-      *error_msg = strdup(error.str().c_str());
-    }
-  }
-  return nullptr;
+  return handle_errors("construct", error_msg,
+                       [&]() { return new bs_model(data, seed); });
 }
 
 void bs_model_destruct(bs_model* m) { delete (m); }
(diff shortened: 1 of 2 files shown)


src/util.hpp
```diff
@@ -29,5 +29,42 @@ inline char* to_csv(const std::vector<std::string>& names) {
...
 }
 
+/**
+ * Convert exception-style error handling into our C output parameter
+ * style. F is always a lambda, so this code gets inlined and the function
+ * call overhead is optimized away.
+ */
+template <typename F>
+[[gnu::always_inline]] [[msvc::forceinline]]
+inline auto handle_errors(const char* name, char** error_msg, F f) {
```
bob-carpenter, line 39: If `F` is a lambda, don't we want to take it in by reference rather than by value?

src/util.hpp
```diff
@@ -29,5 +29,42 @@ inline char* to_csv(const std::vector<std::string>& names) {
...
+  try {
+    return f();
+  } catch (const std::exception& e) {
+    if (error_msg) {
```
bob-carpenter, line 43: Don't we want to print `e.what()` even if `error_msg` is null?  And provide a message no matter what, even if there's no `error_msg` or `e.what()`?

src/util.hpp
```diff
@@ -29,5 +29,42 @@ inline char* to_csv(const std::vector<std::string>& names) {
...
+      std::stringstream error;
+      error << name << "() failed with unknown exception" << std::endl;
+      *error_msg = strdup(error.str().c_str());
+    }
+  }
+
+  // Handle the return value in the failure case, generically
+  using Result = std::invoke_result_t<F>;
+  if constexpr (std::is_same_v<Result, int>) {
```
bob-carpenter, line 58: Should we be explicitly sizing all of our `int` values in C++11 and beyond?  I'm not sure what the best practice is.

src/bridgestan.cpp
```diff
@@ -77,184 +60,66 @@ int bs_param_constrain(const bs_model* m, bool include_tp, bool include_gq,
...
+  });
 }
 
 int bs_log_density_hessian_vector_product(const bs_model* m, bool propto,
                                           bool jacobian,
                                           const double* theta_unc,
                                           const double* v, double* val,
                                           double* Hvp, char** error_msg) {
-  try {
+  return handle_errors("log_density_hessian_vector_product", error_msg, [&]() {
```
bob-carpenter, line 113: if this isn't auto formatted, I'd start the lambda on the next line.

bob-carpenter (commented):
I'll leave this to @roualdes to approve.  Just some comments/questions.

src/util.hpp
WardBrian: Because this function is always being inlined, it’s effectively a kind of macro and this distinction makes no difference on the generated code either way.

Even if this wasn’t the case, taking a lambda by value is fine, and even [recommended as the default choice]( https://johannesugb.github.io/cpu-programming/how-to-pass-lambda-functions-in-C++/#which-variant-shall-i-use) by some

src/util.hpp
WardBrian: That hasn’t been our behavior so far. If the user wanted to see error messages, they shouldn’t have provided us a null pointer for the error message argument

src/util.hpp
bob-carpenter: Interesting.  What's in the body?  Is it just the bound variable pointers and a pointer to code?  I'll read the note.  

If I'm going to do a lot more of this coding I'll need to read the revised Josuttis and Vandevoorde templates book and the Josuttis C++17 book.  What's 1000 pages of dense C++ text?

src/util.hpp
WardBrian: Because the goal of these functions is to be used in FFI, restricting to the “old” C types is the best for compatibility. Julia for example doesn’t define FFI types for the C exact sized integer types

bob-carpenter: Got it.  What a mess all this undefined or partially specified behavior is!

src/util.hpp
WardBrian: Yes, a lambda is basically a struct with elements for the captures and an operator() defined (if it ever actually exists at runtime at all — many times the compiler completely inlines them). It’s generally a good idea to make them cheap to copy, though of course nothing enforces this and you can shoot yourself in the foot if you’re not careful

roualdes (approved):
Looks good to me.

PR #280, opened 2025-06-27
Small updates

WardBrian:
- Link to [polystan](https://github.com/xhep-lab/polystan)
- Update Rust code for new clippy lints in Rust 1.88.0
- Delete unused flag in info(), which was never really set (c.f. https://github.com/stan-dev/cmdstan/issues/1326)

### DIFF
--- a/README.md
+++ b/README.md
@@ -102,6 +102,10 @@ We are aware of the following projects using BridgeStan.
 - https://github.com/UCL/rmcmc
 - https://github.com/CerulloE1996/BayesMVP/
 
+### Other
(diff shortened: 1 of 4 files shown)


bob-carpenter:
> Never heard of PolyChord, but I like this quote from the webpage you link

I dislike it because of the academic-only citation-based license. No open source project with a standard license can depend on it without polluting the license (it's not compatible with BSD, MIT, Apache, or GPL licenses).  So I would urge people to not use such software as it's a dead end.

Issue #202, opened 2024-01-28
Add an R function to convert data list to JSON


avehtari:
In R interface, `StanModel$new()` accepts data as "a JSON string literal, a path to a data file in JSON format ending in “.json”". CmdStanR has handy function `write_stan_json()` https://mc-stan.org/cmdstanr/reference/write_stan_json.html to write R data list to a JSON file compatible with CmdStan (the function does more than direct write of list in json). 
1. It might be useful to have this in bridgestan without dependency on `cmdstanr`
2. It would be awesome to have a function which doesn't require explicitly writing to file, but would just return JSON string literal.

The use case for 2. is when there is a need to instantiate the model with new data with minimal overhead. In the following code snippet `sd` is a data list, and we update part of the list, call `new`, and use log density, gradient and Hessian. Based on timing experiments, this is slightly faster than using ``write_stan_json()` to write to  a file and passing the file path to `new()`
```
sd$tau <- draws_sd[s, ]
model <- StanModel$new(model_so, to_stan_json(sd), SEED)
```

Another use case is when using BridgeStan with brms generated models and data
```
model_so <- compile_model(write_stan_file(stancode(brms_fit)))
model <- StanModel$new(model_so, to_stan_json(stan_data(brms_fit)), SEED)
```

`to_stan_json()` can be modified from `write_stan_json` by removing references to file (the code is using BSD-3 license).
```
to_stan_json <- function(data, always_decimal = FALSE) {
  if (!is.list(data)) {
    stop("'data' must be a list.", call. = FALSE)
  }

  data_names <- names(data)
  if (length(data) > 0 &&
      (length(data_names) == 0 ||
       length(data_names) != sum(nzchar(data_names)))) {
    stop("All elements in 'data' list must have names.", call. = FALSE)

  }
  if (anyDuplicated(data_names) != 0) {
    stop("Duplicate names not allowed in 'data'.", call. = FALSE)
  }

  for (var_name in data_names) {
    var <- data[[var_name]]
    if (!(is.numeric(var) || is.factor(var) || is.logical(var) ||
          is.data.frame(var) || is.list(var))) {
      stop("Variable '", var_name, "' is of invalid type.", call. = FALSE)
    }
    if (anyNA(var)) {
      stop("Variable '", var_name, "' has NA values.", call. = FALSE)
    }

    if (is.table(var)) {
      var <- unclass(var)
    } else if (is.logical(var)) {
      mode(var) <- "integer"
[...]

PR #283, opened 2025-07-01
Allow passing dictionaries to Python functions using stanio

WardBrian:
Requested in https://github.com/roualdes/bridgestan/issues/202#issuecomment-1915032292

This uses the [stanio](https://github.com/stan-dev/stanio) which is a small library also used by CmdStanPy to allow passing of Python dictionaries in the couple places we previously required raw JSON strings

### DIFF
--- a/python/bridgestan/model.py
+++ b/python/bridgestan/model.py
@@ -2,11 +2,12 @@ import ctypes
 import warnings
 from os import PathLike, fspath
 from pathlib import Path
-from typing import List, Optional, Tuple, Union
+from typing import Any, Mapping, List, Optional, Tuple, Union
 
+import dllist
+import stanio
 import numpy as np
 import numpy.typing as npt
-from dllist import dllist
 from numpy.ctypeslib import ndpointer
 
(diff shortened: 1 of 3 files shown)


roualdes:
Let me know when this is ready for review and I'll look it over.

WardBrian:
Should be ready now!

python/bridgestan/model.py
```diff
@@ -36,9 +37,10 @@ def from_param(cls, obj):
 
 FloatArray = Union[
     npt.NDArray[np.float64],
-    ctypes.POINTER(ctypes.c_double),
+    "ctypes._Pointer[ctypes.c_double]",
```
roualdes, line 40: Why the change here?

WardBrian: Two parter:

ctypes.POINTER is a value, not a type (like saying we had a Union of 1, vs int). My bad

ctypes._Pointer is the corresponding type, but it is not a real generic at runtime, hence why it is quotes

roualdes (approved):

PR #288, opened 2025-09-03
Update documentation url to roualdes.us

WardBrian:
Save our users one redirect (the old links do still work, for what it's worth)


WardBrian:
> Should we also change the URLs in?
    rust/Cargo.toml
    rust/README.md
    docs/conf.py

Those were in this PR. I pushed a separate fix to conf.py directly to main to fix the website build

PR #294, opened 2025-10-08
param_unc_names takes const self

aseyboldt:
`param_unc_names` takes a mutable reference in the rust wrapper, when a const reference is enough.


WardBrian (approved):
Thanks!

PR #304, opened 2025-11-13
Test prerelease of Julia 1.12.2

WardBrian:


WardBrian:
Huh -- the [announcement blog](https://discourse.julialang.org/t/julia-v1-12-2-testing-period/133847) claimed that `1.12-nightly` would work, but I guess not. I'll try again when it actually releases

PR #320, opened 2026-05-28
Add method for initializing parameters

### LINKED ISSUE #319: New method: initialize
This would basically serve as a wrapper around [`stan::services::utils::initialize`](https://github.com/stan-dev/stan/blob/develop/src/stan/services/util/initialize.hpp).

We currently provide `bs_param_unconstrain_json`, but this is insufficient for the use case of specifying _some_ of the parameters and allowing others to be randomly initialized.

The signature would be 

```C
BS_PUBLIC int bs_param_initialize(const bs_model* m, const char* json,
                                  bs_rng* rng, double init_radius, 
                                  bool jacobian, double* theta_unc, char** error_msg);
```

Note that this function does perform gradient evals to ensure that the random values lead to a non-infinite log density and gradient

WardBrian:
Closes #319 

All I had time for today was the C++ side of this, I will circle back next week with the various client wrappers.


WardBrian:
Seems like https://github.com/JuliaLang/julia/issues/59931 is back, possibly due to https://github.com/actions/runner-images/issues/14017?

aseyboldt:
If there was an option to also reject points with zero gradient, then nutpie could also use the same function. :-)

WardBrian:
I suppose nutpie could use the same function anyway with its own loop to check the non-zero condition. Outside of a trivial case of a standard normal initialized at 0, I doubt there are that many in-the-wild models that would initialize in such a place, or am I underestimating something?

aseyboldt:
True, I can just set the retries to zero or one, and call it in a loop myself...
It does happen in miss-specified models from time to time.

bob-carpenter:
To some extent, nutpie struggles with initialization if gradients are close to zero, which is easy to get with a randomly initialized 100-dimensional normal.  I included some additive regularization so that the mass matrix diagonal entries don't get too large or too small, where "too" is completely ill-defined.  E.g.,

```R
> min(abs(rnorm(100, 0, 1)))
[1] 0.008291986

> min(abs(rnorm(100, 0, 1)))
[1] 0.004029361

> min(abs(rnorm(100, 0, 1)))
[1] 0.001589286

> min(abs(rnorm(100, 0, 1)))
[1] 0.00323525
```

The gradient is just negative the value for a standard normal.  We don't really want to initialize mass matrix entries for independent standard normal marginals in the (0.001, 0.01) range.

I'm thinking that what we should be doing is taking all of our initializations and taking the average of their mass matrices (probably regular rather than geometric).

WardBrian:
This is ready for review @roualdes @bob-carpenter 

The only oddity is that I am bumping the Stan submodule to a non-released version, to allow us to specify the maximum number of tries. The latest release was so recent that this is more or less the only code change in the submodule, so I think this is fine to do, just wanted to flag it. The timing out Julia windows test is unrelated, I'm debugging in #321

docs/languages/_r/StanModel.md
```diff
@@ -342,6 +342,44 @@ The unconstrained parameters of the model.
 
  
 
+##### Method `param_initialize()`
+
+Initialize a point in the unconstrained space, using the specified values from JSON and randomizing the others. The result will be checked to ensure they achieve a finite log density and gradient. The JSON is expected to be in the [JSON Format for CmdStan](https://mc-stan.org/docs/cmdstan-guide/json_apdx.html).
```
bob-carpenter: Can you make the lines no more than 80 characters wide so they're easier to review? 

I think we either need to say more about what "will be checked" means or defer that until later when we talk about retries.

docs/languages/_r/StanModel.md
```diff
@@ -342,6 +342,44 @@ The unconstrained parameters of the model.
...
+  jacobian = TRUE
+)
+```
+
+ 
+
+###### Arguments
+
+- **`rng`**: The source of randomness for the unspecified parameters.
+- **`json`**: Character vector containing a string representation of JSON data.
```
bob-carpenter, line 370: Shouldn't this be a dictionary rather than JSON?  JSON is just a serialized dictionary, so why go through the serialization/deserialization for an argument?

docs/languages/_r/StanModel.md
```diff
@@ -342,6 +342,44 @@ The unconstrained parameters of the model.
...
+- **`init_radius`**: The parameters not provided will be drawn uniformly from `[-init_range, init_range]` on the unconstrained scale.
+- **`max_tries`**: How many attempts should be made to find a point with finite log density.
```
bob-carpenter: This should say "Maximum number of random initializations considered to find an initialization with finite log density."  It's not technically how many attempts, because fewer may be attempted than this number.

docs/languages/_r/StanModel.md
```diff
@@ -342,6 +342,44 @@ The unconstrained parameters of the model.
...
+- **`jacobian`**: If `TRUE`, include change of variables terms for constrained parameters when checking for finiteness.
```
bob-carpenter: I had to think hard about this one.

First, we're not checking the change of variables themselves for infinite variables, we're talking about including the Jacobian from the transform.  The use of "terms" clarifies somewhat, but we need to at least say "log density terms" to make it clearer.

Second, we need to provide a suggestion somewhere that the only time this would be set to `FALSE` would be for doing optimization without the Jacobian term (i.e. (possibly penalized) maximum likelihood).

docs/languages/_r/StanModel.md
```diff
@@ -342,6 +342,44 @@ The unconstrained parameters of the model.
...
+
+ 
+
+###### Returns
+
+The unconstrained parameters of the model.
```
bob-carpenter, line 379: What happens if it doesn't find a finite log density?

docs/languages/_rust/bridgestan/lib.md
```diff
@@ -374,6 +374,21 @@
   *Panics* if `include_gq` is set but no random number generator is provided.
```
bob-carpenter, line 374: What does "panics" mean here?   Is that some kind of Rust lingo for throwing an exception or an assert?

docs/languages/_rust/bridgestan/lib.md
```diff
@@ -374,6 +374,21 @@
...
+
+  :::
+  Initialize a point in the unconstrained space, using the specified values
+  from JSON and randomizing the others.
+  
+  The JSON is expected to be in the
+  [JSON Format for CmdStan](https://mc-stan.org/docs/cmdstan-guide/json.html).
+  The result will be checked to ensure they achieve a finite
```
bob-carpenter, line 388: they achieve -> the initial point has

docs/languages/_rust/bridgestan/lib.md
```diff
@@ -411,7 +426,7 @@
 ::::{rust:function} bridgestan::Model::param_unc_names
 :index: -1
 :vis: pub
-:layout: [{"type":"keyword","value":"fn"},{"type":"space"},{"type":"name","value":"param_unc_names"},{"type":"punctuation","value":"("},{"type":"punctuation","value":"&"},{"type":"keyword","value":"mut"},{"type":"space"},{"type":"keyword","value":"self"},{"type":"punctuation","value":")"},{"type":"space"},{"type":"returns"},{"type":"space"},{"type":"punctuation","value":"&"},{"type":"link","value":"str","target":"str"}]
+:layout: [{"type":"keyword","value":"fn"},{"type":"space"},{"type":"name","value":"param_unc_names"},{"type":"punctuation","value":"("},{"type":"punctuation","value":"&"},{"type":"keyword","value":"self"},{"type":"punctuation","value":")"},{"type":"space"},{"type":"returns"},{"type":"space"},{"type":"punctuation","value":"&"},{"type":"link","value":"str","target":"str"}]
```
bob-carpenter, line 429: This is some Rust/markdown magic I don't understand!

docs/languages/julia.md
```diff
@@ -261,6 +261,24 @@ This allocates new memory for the output each call. See [`param_unconstrain_json
...
+<a id='BridgeStan.param_initialize' href='#BridgeStan.param_initialize'>#</a>
+**`BridgeStan.param_initialize`** &mdash; *Function*.
+
+
+
+```julia
+param_initialize(sm, rng, theta="{}"; init_radius=2.0, max_tries=100, jacobian=true)
+```
+
+Similar to [`param_unconstrain_json`](julia.md#BridgeStan.param_unconstrain_json) but specifically allows for incomplete specifications of the parameters. Any parameter not specified in the provided JSON will be randomized. The result will be checked for a finite log density value, and re-tried up to the specified number of times.
```
bob-carpenter: Line length.

bob-carpenter: What happens if all the retries fail?  

re-tried -> retried

docs/languages/julia.md
```diff
@@ -261,6 +261,24 @@ This allocates new memory for the output each call. See [`param_unconstrain_json
...
+
+The JSON is expected to be in the [JSON Format for CmdStan](https://mc-stan.org/docs/cmdstan-guide/json_apdx.html).
+
+This allocates new memory for the output each call. See [`param_initialize!`](julia.md#BridgeStan.param_initialize!) for a version which allows re-using existing memory.
```
bob-carpenter: re-using -> reusing

(the spell checkers want to hyphenate things that are conventionally written without hyphens)

docs/languages/julia.md
```diff
@@ -469,6 +487,24 @@ The result is stored in the vector `out`, and a reference is returned. See [`par
...
+<a id='BridgeStan.param_initialize!' href='#BridgeStan.param_initialize!'>#</a>
+**`BridgeStan.param_initialize!`** &mdash; *Function*.
+
+
+
+```julia
+param_initialize!(sm, rng, out, theta="{}"; init_radius=2.0, max_tries=100, jacobian=true)
+```
+
+Similar to [`param_unconstrain_json!`](julia.md#BridgeStan.param_unconstrain_json!) but specifically allows for incomplete specifications of the parameters. Any parameter not specified in the provided JSON will be randomized. The result will be checked for a finite log density value, and re-tried up to the specified number of times.
```
bob-carpenter: specifically -> additionally -or- [delete]

randomized -> initialized randomly with `uniform((-init_radius, init_radius))`

re-tried -> retried

Say what happens if all the retries fail.

julia/src/model.jl
```diff
@@ -449,6 +449,89 @@ function param_unconstrain_json(sm::StanModel, theta::String)
...
 end
 
+
+"""
+    param_initialize!(sm, rng, out, theta=\"{}\"; init_radius=2.0, max_tries=100, jacobian=true)
+
+Similar to [`param_unconstrain_json!`](@ref) but specifically allows for
+incomplete specifications of the parameters. Any parameter not specified
+in the provided JSON will be randomized. The result will be checked for a
+finite log density value, and re-tried up to the specified number of times.
```
bob-carpenter: I hate to keep saying this again and again, but I still don't know what happens if the retries all fail.

julia/src/model.jl
```diff
@@ -449,6 +449,89 @@ function param_unconstrain_json(sm::StanModel, theta::String)
...
+        error(handle_error(sm.lib, err, "param_initialize"))
+    end
+    out
+end
+
+"""
+    param_initialize(sm, rng, theta=\"{}\"; init_radius=2.0, max_tries=100, jacobian=true)
+
+
+Similar to [`param_unconstrain_json`](@ref) but specifically allows for
```
bob-carpenter: same comment as before on "specifically", "re-tried", "randomized", and what happens if all check fails.

python/bridgestan/model.py
```diff
@@ -543,6 +556,65 @@ def param_unconstrain_json(
...
+        init_radius: float = 2.0,
+        max_tries: int = 100,
+        jacobian: bool = True,
+        out: Optional[FloatArray] = None,
+    ) -> FloatArray:
+        """
+        Return an array of the unconstrained parameters derived from the
+        specified JSON formatted data, allowing for incomplete specifications.
+        If not all parameters are specified in the JSON, the remainder will be
+        randomized. The result will be checked to ensure they achieve a finite
```
bob-carpenter: This needs to say how they're randomized, at least to the extent of "uniformly in an interval on the unconstrained scale" or something like that.  As is, this is only documented on `init_radius`, which seems a bit obscure.

python/bridgestan/model.py
```diff
@@ -543,6 +556,65 @@ def param_unconstrain_json(
...
+        log density and gradient.
+
+        The JSON is expected to be in the `JSON Format for CmdStan <https://mc-stan.org/docs/cmdstan-guide/json.html>`__.
+
+        :param rng: The source of randomness for the unspecified parameters.
+        :param theta_json: The JSON encoded constrained parameters or a dictionary
+            which will be converted to a JSON string.
+        :param init_radius: The parameters not provided will be drawn uniformly
+            from ``[-init_range, init_range]`` on the unconstrained scale.
```
bob-carpenter: `[...]` -> `[...)` for uniform RNGS---they can generate their lower bounds but not their upper bounds.  We probably have this inconsistently noted all over our documentation.  I just looked up what the Boost RNGs do and what the built-in C++ RNGs do.

python/bridgestan/model.py
```diff
@@ -543,6 +556,65 @@ def param_unconstrain_json(
...
+        :param max_tries: How many attempts should be made to find a point with finite
+            log density.
+        :param jacobian: ``True`` if change-of-variables terms for constrained
```
bob-carpenter, line 586: I would say Jacobian terms here because of the possible confusion with checking the transformed variables for finiteness.

R/man/StanModel.Rd
```diff
@@ -286,6 +287,47 @@ The unconstrained parameters of the model.
 }
 }
 \if{html}{\out{<hr>}}
+\if{html}{\out{<a id="method-StanModel-param_initialize"></a>}}
+\if{latex}{\out{\hypertarget{method-StanModel-param_initialize}{}}}
+\subsection{Method \code{param_initialize()}}{
+Initialize a point in the unconstrained space, using the specified values
+from JSON and randomizing the others.
+The result will be checked to ensure they achieve a finite
```
bob-carpenter: I wish I could just make a blanket statement asking what "checked" means here!

R/man/StanModel.Rd
```diff
@@ -286,6 +287,47 @@ The unconstrained parameters of the model.
...
+  jacobian = TRUE
+)}\if{html}{\out{</div>}}
+}
+
+\subsection{Arguments}{
+\if{html}{\out{<div class="arguments">}}
+\describe{
+\item{\code{rng}}{The source of randomness for the unspecified parameters.}
+
+\item{\code{json}}{Character vector containing a string representation of JSON data.}
```
bob-carpenter, line 317: Same question about why this isn't just a list rather than JSON.

R/man/StanModel.Rd
```diff
@@ -286,6 +287,47 @@ The unconstrained parameters of the model.
...
+
+\item{\code{init_radius}}{The parameters not provided will be drawn uniformly
+from \verb{[-init_range, init_range]} on the unconstrained scale.}
```
bob-carpenter: [)

R/man/StanModel.Rd
```diff
@@ -286,6 +287,47 @@ The unconstrained parameters of the model.
...
+
+\item{\code{max_tries}}{How many attempts should be made to find a point with finite
+log density.}
+
+\item{\code{jacobian}}{If \code{TRUE}, include change of variables terms for constrained
```
bob-carpenter, line 325: Same comment as elsewhere.

R/man/StanModel.Rd
```diff
@@ -286,6 +287,47 @@ The unconstrained parameters of the model.
...
+parameters when checking for finiteness.}
+}
+\if{html}{\out{</div>}}
+}
+\subsection{Returns}{
+The unconstrained parameters of the model.
+}
+}
```
bob-carpenter, line 333: I want to know what happens if all tests for finiteness fail.

R/R/bridgestan.R
```diff
@@ -317,6 +317,53 @@ StanModel <- R6::R6Class(
       vars$theta_unc
     },
     #' @description
+    #' Initialize a point in the unconstrained space, using the specified values
+    #' from JSON and randomizing the others.
+    #' The result will be checked to ensure they achieve a finite
+    #' log density and gradient.
```
bob-carpenter: Same comments as elsewhere.  This is getting tedious because of all the cut-and-paste, because I just have the exact same comments everywhere.

R/R/bridgestan.R
```diff
@@ -317,6 +317,53 @@ StanModel <- R6::R6Class(
...
+    #' The JSON is expected to be in the [JSON Format for CmdStan](https://mc-stan.org/docs/cmdstan-guide/json_apdx.html).
+    #'
+    #' @param rng The source of randomness for the unspecified parameters.
+    #' @param json Character vector containing a string representation of JSON data.
+    #' @param init_radius The parameters not provided will be drawn uniformly
+    #' from `[-init_range, init_range]` on the unconstrained scale.
```
bob-carpenter: [)

R/R/bridgestan.R
```diff
@@ -317,6 +317,53 @@ StanModel <- R6::R6Class(
...
+    #' @param max_tries How many attempts should be made to find a point with finite
+    #' log density.
+    #' @param jacobian If `TRUE`, include change of variables terms for constrained
```
bob-carpenter, line 336: Ditto.

R/R/bridgestan.R
```diff
@@ -317,6 +317,53 @@ StanModel <- R6::R6Class(
...
+          private$lib_name,
+          vars$err_msg,
+          vars$err_ptr,
+          "param_initialize"
+        ))
+      }
+      vars$theta_unc
+    },
+    #' @description
     #' Return the log density of the specified unconstrained parameters.
```
bob-carpenter, line 371: Say what happens if finiteness checks fail.

rust/src/bs_safe.rs
```diff
@@ -705,6 +705,47 @@ impl<T: Borrow<StanLibrary>> Model<T> {
             Err(BridgeStanError::EvaluationFailed(err.message()))
         }
     }
+
+    /// Initialize a point in the unconstrained space, using the specified values
+    /// from JSON and randomizing the others.
```
bob-carpenter, line 710: I'd just do a global search for "randomizing" and make this a little clearer.  Not a big deal and I realize I'm just asking you to change a dozen cut-and-paste versions.  This project is painful in requiring duplicated doc across so many languages.

src/bridgestan.h
```diff
@@ -214,13 +217,47 @@ BS_PUBLIC int bs_param_unconstrain(const bs_model* m, const double* theta,
...
+ * is an error. This must later be freed by calling \link bs_free_error_msg()
+ * \endlink.
  * @return code 0 if successful and code -1 if there is an exception
  * in the underlying Stan code
  */
 BS_PUBLIC int bs_param_unconstrain_json(const bs_model* m, const char* json,
                                         double* theta_unc, char** error_msg);
 
+/**
+ * Similar to \link bs_param_unconstrain_json() \endlink, but specifically
```
bob-carpenter, line 229: specifically -> ditto, randomized.

Rather than "an error will be indicated", say it returns the failure return code.

src/bridgestan.h
```diff
@@ -214,13 +217,47 @@ BS_PUBLIC int bs_param_unconstrain(const bs_model* m, const double* theta,
...
+ * The JSON is expected to be in the
+ * <a href="https://mc-stan.org/docs/cmdstan-guide/json.html">JSON Format for
+ * CmdStan</a>.
+ *
+ * @param[in] m pointer to model structure
+ * @param[in] json JSON-encoded constrained parameters
+ * @param[in] rng Random number generator to use for the parameters not provided
+ * in the JSON.
+ * @param[in] init_radius The parameters not provided will be drawn uniformly
+ * from `[-init_range, init_range]` on the unconstrained scale.
```
bob-carpenter: [). this may be getting a bit pedantic :-)

src/bridgestan.h
```diff
@@ -214,13 +217,47 @@ BS_PUBLIC int bs_param_unconstrain(const bs_model* m, const double* theta,
...
+ * @param[in] max_tries maximum number of attempts at random initialization
+ * @param[in] jacobian whether to use the jacobian when calculating if the log
+ * density is finite.
+ * @param[out] theta_unc sequence of unconstrained parameters
+ * @param[out] error_msg a pointer to a string that will be allocated if there
```
bob-carpenter, line 251: Is failing all retries an error?

src/bridgestan.h
```diff
@@ -214,13 +217,47 @@ BS_PUBLIC int bs_param_unconstrain(const bs_model* m, const double* theta,
...
+ * is an error. This must later be freed by calling \link bs_free_error_msg()
```
bob-carpenter, line 252: Why allocate a string rather than accept something like an error stream that won't have memory implications for this function?

src/bridgestanR.cpp
```diff
@@ -55,6 +55,15 @@ void bs_param_unconstrain_json_R(bs_model** model, char const** json,
...
   *err_ptr = static_cast<void*>(*err_msg);
 }
+void bs_param_initialize_R(bs_model** model, char const** json, bs_rng** rng,
+                           double* init_radius, int* max_tries, int* jacobian,
+                           double* theta_unc, int* return_code, char** err_msg,
+                           void** err_ptr) {
+  *return_code
+      = bs_param_initialize(*model, *json, *rng, *init_radius, *max_tries,
+                            (*jacobian != 0), theta_unc, err_msg);
+  *err_ptr = static_cast<void*>(*err_msg);
```
bob-carpenter, line 65: Question:

Why are we returning both `err_msg` and `err_ptr` when `err_msg` is recoverable from `err_ptr` (I think)?

bob-carpenter (changes requested):
I had some genuine questions, but overall this looks great.  The change requests are all around clarity in the descriptions---sorry those got cut-and-paste across all interfaces, but it's pretty much the same changes suggested everywhere.

If you think clarifying RNG ranges to be `[ , )` is too much, I'd be OK skipping that.

src/bridgestanR.cpp
WardBrian: Trying too hard to understand these R wrappers is a path to madness, but the short answer is that R's `.C` ffi function copies strings on both ends of the call. So unless we pass the value of the pointer some other way like this, we'd have no way to free the original string

src/bridgestan.h
WardBrian: This was also your instinct back in 2023 when we first added error handling to the library (https://github.com/roualdes/bridgestan/issues/90#issuecomment-1483376151)!

The short summary is that it's nice to keep the C interface as simple as possible in terms of the types we need to pass. And there are not really good, easy-to-instantiate options for a stream at the C level. The simplest would probably be if we took a function pointer, but that would be hard (if not impossible?) in R, and would be somewhat tricky to write in other languages to ensure e.g. the action taken on error is thread-safe.

The  interfaces in other languages all encapsulate the `char**` nicely in a way that means most user's don't need to worry. And we have it such that NULL is a valid option, which just makes it so the error message isn't allocated anywhere.

As an aside, I learned about a year after we released 2.0 of BridgeStan with these error messages that this style had been invented before by Stefanus Du Toit under the name "Hourglass Interfaces". His [CppCon 2014 talk](https://www.youtube.com/watch?v=PVYdHDm0q6Y) slides show nearly this exact style on slides 30-33. The only difference is he recommends wrapping the error in a opaque struct and forcing the caller to use a function to get the message out (which also allows you to contain more than just a message in an error). This is the style I used [in tinystan](https://github.com/WardBrian/tinystan/blob/main/src/tinystan.h#L337-L367), and would probably switch to here if we didn't care about breaking changes

It's almost eerie looking at the slides from that talk, because someone probably could have watched it and come up with _very nearly exactly the same API_ for BridgeStan, just ~9 years before @roualdes ultimately did!

docs/languages/_r/StanModel.md
WardBrian: These markdown files are automatically generated from the comments in the R code (which is line-wrapped). I can't control them directly

docs/languages/_rust/bridgestan/lib.md
WardBrian: Yes, a [panic](https://doc.rust-lang.org/std/macro.panic.html) is basically an exception in Rust. The connotation is different, because they're not used for control flow or 'normal' error handling like exceptions in C++, but instead to indicate that you've written buggy code which has done something inherently incorrect

docs/languages/_rust/bridgestan/lib.md
WardBrian: Yeah, this is another automatically generated file. All this metadata is just so things are highlighted/clickable as expected in the rendered HTML, I believe

python/bridgestan/model.py
WardBrian: Huh, it looks like [Stan's own documentation](https://mc-stan.org/docs/cmdstan-guide/command_line_options.html#initialize-model-parameters-argument) has this same error, which is probably where I picked it up. As far as I can tell, it's been wrong since the first time it was documented

WardBrian: [Another place](https://mc-stan.org/docs/reference-manual/execution.html#random-initial-values) says `(x, x)` instead, so we've got both the incorrect answers written down and nowhere the right one!

src/bridgestan.h
bob-carpenter: I'm usually consistent enough in these kinds of things that it sounds like I'm quoting myself needlessly.  But I forget what I said.

Always nice to see something you independently discover turn out to be "real."  It's the usual experience during the first couple years in a field when you realize someone has already had all the obviously good ideas, but you just don't know about them yet.

We can do it the way you like it for Walnuts, then :-).

docs/languages/_r/StanModel.md
WardBrian: We don't currently have a solution to reliably convert R lists to JSON, so we force the user to do it before calling. This is https://github.com/roualdes/bridgestan/issues/202

src/bridgestan.h
WardBrian: Yes, any nonzero return code will set a string here (though sometimes it might just be 'unknown error' if an exception we don't know about got caught)

bob-carpenter (approved):
Thanks for answering all my questions.  I still don't like "achieves" or "change of variables terms", but I guess I can't have everything.

PR #321, opened 2026-06-03
Julia debug

WardBrian:
Rather than continuing to pollute #320 I am using this as a scratch pad

### DIFF
--- a/.github/workflows/main.yaml
+++ b/.github/workflows/main.yaml
@@ -8,7 +8,7 @@ on:
   workflow_dispatch: {}
 
 env:
-  CACHE_VERSION: 1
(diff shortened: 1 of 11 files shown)


WardBrian:
This is looking more and more like a runtime library conflict, since adding a ucrt mingw build to the path broke windows-2022, which was previously working. I can't figure out what to do about it, though...

WardBrian:
Forcing msvcrt also broke windows-2022, so maybe it’s a version specific thing rather than a runtime. To try tomorrow

WardBrian:
Using an older version allows some progress/output to be made, but does still eventually hang. Weird!

PR #325, opened 2026-06-12
Allow NULL for json in bs_param_initialize

WardBrian:
Re-uses some code we had for the data in `bs_model_construct`, which has already allowed NULL

### DIFF
--- a/python/bridgestan/model.py
+++ b/python/bridgestan/model.py
@@ -559,7 +559,7 @@ class StanModel:
     def param_initialize(
         self,
         rng: "StanRNG",
-        theta_json: Union[str, Mapping[str, Any]] = "{}",
+        theta_json: Union[str, Mapping[str, Any], None] = None,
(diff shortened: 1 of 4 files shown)


roualdes (changes requested):
The changes look good. 

I've got a question about a non-change(?). Did we lose our test on an empty dictionary, e.g. `python/test/test_stanmodel.py` inside `def test_param_initialize()`?

WardBrian:
The behavior of the tests definitely did change slightly with the new default for the argument, I've added more explicit testing of the empty case

roualdes (approved):
Looks great.  Thanks.

WardBrian:
@roualdes FYI I am planning on pressing 'release' on 2.9 soon, with this new function being the primary change

roualdes:
Sounds good. I've already started using it. Thanks.

PR #331, opened 2026-07-28
Python: move windows dll setup to after possible compilation

WardBrian:
If someone installs bridgestan for the first time and immediately calls `bridgestan.StanModel` with a `.stan` file, the `windows_dll_setup` function will be called _before_ TBB was ever built, which causes it to fail (observed in https://github.com/flatironinstitute/walnuts/pull/78). Moving it below the branch which can call compile fixes this, and was where the similar function was already called in the other languages

### DIFF
--- a/python/bridgestan/model.py
+++ b/python/bridgestan/model.py
@@ -134,13 +134,12 @@ class StanModel:
             else:
                 data = stanio.dump_stan_json(data)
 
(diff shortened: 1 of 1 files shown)


roualdes (approved):
Seems a simple enough fix for a positive outcome.

PR #341, opened 2026-09-07
Quote Make arguments when compiling models from R interface

### LINKED ISSUE #340: R interface does not quote BridgeStan path argument to Make in `compile_model`
On the Windows machine of a colleague, @Sam0287, when trying to compile a Stan model using BridgeStan in R we experienced an error

> Error in compile_model(lib, stanc_args, make_args) : Compilation failed with error code 2 output: make: Entering directory '/c/Users/Sam/OneDrive' make: *** No rule to make target 'University'. Stop. make: Leaving directory '/c/Users/Sam/OneDrive'

The last `OneDrive` segment in the path in the error message is only part of that segment, up to a space character, with BridgeStan having defaulted to download the BridgeStan library to the default user directory which on this machine mapped to a institutional OneDrive synced directory with the institution name, and spaces, included within the directory path.

We managed to resolve the error by manually setting the path to the BridgeStan directory to a temporary directory path (without any spaces)  using `bridgestan::set_bridgestan_path` and extracting an archived release of the BridgeStan library there. I believe therefore the issue is with spaces in the path not being quoted (or escaped) when passed through to the Make command used to compile the model, constructed in the lines

https://github.com/roualdes/bridgestan/blob/ddf067ea72007ef569a8ba3b17b297a8dac38bae/R/R/compile.R#L125-L137

It appears that `get_bridgestan_path` returns an unquoted path, hence if it contains spaces, the call to make in the `system2` call using the constructed `flags` arguments will interpret the path components split on spaces as separate command line arguments.

It looks like the potential for similar issues in `stancflags` are being guarded against by manually quoting the constructed string in `flags` with `paste0("STANCFLAGS=\"", stancflags, "\"")`. 

From a bit of digging I think the [built-in `shQuote` function in R](https://www.rdocumentation.org/packages/base/versions/3.6.2/topics/shQuote) would offer a more system agnostic approach to quoting both the `stancflags` and output of `get_bridgestan_path` - so changing the statement constructing `flags` string to something like

```R
flags <- c(
  paste("-C", shQuote(get_bridgestan_path())),
  make_args,
  paste0("STANCFLAGS=", shQuote(stancflags)),
  output
)
```

I think would resolve this issue and make the `stancflags` handling more robust.

@roualdes If this seems like a reasonable fix I can open a pull-request with these changes if that is helpful, or with just the changes applied to `get_bridgestan_path` output handling if it's better to leave the current `stancflags` handling as is.

matt-graham:
Potentially resolves #340 

Uses base R `shQuote` function to quote output of `get_bridgestan_path` and `stancflags` passed as arguments to Make in calls to `compile_model` from R interface, to avoid possible issues with for example spaces in paths being otherwise interpreted as an argument separator.

Also adds a test that temporarily sets `BRIDGESTAN` environment variable to point to a temporary directory containing spaces in the path (and with BridgeStan source copied to that directory), and checks whether a model can be compiled successfully. Before changes to `compile.R` in this PR this test fail while it is passing locally for me after changes.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -72,17 +72,17 @@ endif
 %.hpp : %.stan $(STANC)
 	@echo ''
 	@echo '--- Translating Stan model to C++ code ---'
-	$(STANC) $(STANCFLAGS) --o=$(subst  \,/,$@) $(subst  \,/,$<)
+	$(STANC) $(STANCFLAGS) --o="$(subst \,/,$@)" "$(subst \,/,$<)"
 
 %.o : %.hpp $(USER_HEADER)
 	@echo ''
 	@echo '--- Compiling C++ code ---'
-	$(COMPILE.cpp) $(USER_INCLUDE) -x c++ -o $(subst  \,/,$*).o $(subst \,/,$<)
+	$(COMPILE.cpp) $(USER_INCLUDE) -x c++ -o "$(subst \\,/,$*).o" "$(subst \\,/,$<)"
 
 %_model.so : %.o $(BRIDGE_O) $(SUNDIALS_TARGETS) $(MPI_TARGETS) $(TBB_TARGETS)
 	@echo ''
 	@echo '--- Linking C++ code ---'
-	$(LINK.cpp) -shared -lm -o $(patsubst %.o, %_model.so, $(subst \,/,$<)) $(subst \,/,$*.o) $(BRIDGE_O) $(LDLIBS) $(SUNDIALS_TARGETS) $(MPI_TARGETS) $(TBB_TARGETS)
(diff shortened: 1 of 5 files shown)


WardBrian (approved):
Thanks!

We should probably look into similar functionality in python/julia, which I'll do this week before merging this

WardBrian:
@matt-graham when writing the equivalent test in Python I wrote it in a slightly different style that tried to build the test model from the temporary copy. This revealed a related issue (which is also encountered by R) when there is a space in the _path to the model_, which is something stemming from our makefiles ~with a far less obvious fix.~

Edit: I think it is actually relatively simple to update our makefiles

